PR #8, opened 2023-12-13
changing arb goerli to arb sepolia on examples

GreatSoshiant:
Because we just have Arb Sepolia and Sepolia chains defined for rollup creation, Arb Goerli txs would be rejected so these script can't be executed. I've decided to change it to Arb Sepolia and not adding Arb Goerli to the chain list cause Arb Goerli is going to be depreciated.


hkalodner (approved):
LGTM

PR #11, opened 2023-12-14
feat: add upgrade executor encode function data util

spsjvc:
Instead of having `executeCallEncodeFunctionData` it's now more generic so it can be used for any function on `UpgradeExecutor`.


fionnachan (approved):
LGTM

PR #26, opened 2023-12-20
fix: properly set callvalue in create rollup

spsjvc:
The function `createRollupGetCallValue` was being called without applying defaults, which means `params.deployFactoriesToL2` was set to `false` (even though are defaults set it to `true`) so the call value was always 0. 

Now we properly apply defaults to input params first.

I also added one extra assertion to make sure everything was executed properly and the logs can be successfully decoded.

The test failing in CI should be ignored as it's not set up properly, and should be tried out locally.


TucksonDev (approved):
LGTM. CI is failing, but it was tested with a custom script.

PR #28, opened 2023-12-21
ci: update nitro-testnode params

spsjvc:
Fixes argument `ref` to `nitro-testnode-ref` and adds the following flags:

```
// RollupCreator on L1 is deployed by default
--tokenbridge // this will make sure we have the TokenBridgeCreator deployed on L1
--l3node // this will make sure we have RollupCreator deployed on L2
--l3-token-bridge // this will make sure we have TokenBridgeCreator deployed on L2
```


TucksonDev (approved):
🚀 🚀 🚀 🚀 🚀

PR #29, opened 2023-12-21
feat: add upgradeExecutor management functions

### LINKED ISSUE #13: Add "upgradeExecutorFetchPrivilegedAccounts" util to upgrade executor

TucksonDev:
This PR adds functions to manage accounts with the Executor role in an UpgradeExecutor. It also includes a function to fetch all privileged accounts from an UpgradeExecutor. The PR includes an integration test for adding/removing executor accounts, and a unit test for the fetch function. This PR also includes an example script to set a new executor account in both the parent and Orbit chains.

New exposed functions:
- upgradeExecutorEncodeFunctionData
- upgradeExecutorFetchPrivilegedAccounts
- upgradeExecutorPrepareAddExecutorTransactionRequest
- upgradeExecutorPrepareRemoveExecutorTransactionRequest

 Closes #13
 Fixes INT-123


src/upgradeExecutorPrepareAddExecutorTransactionRequest.ts
```diff
@@ -0,0 +1,59 @@
...
+  upgradeExecutorAddress: Address;
+  executorAccountAddress: Address;
+  publicClient: PublicClient;
+};
+
+export async function upgradeExecutorPrepareAddExecutorTransactionRequest({
+  account,
+  upgradeExecutorAddress,
+  executorAccountAddress,
+  publicClient,
```
TucksonDev, line 19: The same function is used when adding an account to the UpgradeExecutor of the parent chain, and the Orbit chain.

src/upgradeExecutorPrepareAddExecutorTransactionRequest.ts
```diff
@@ -0,0 +1,59 @@
...
+    abi: upgradeExecutor.abi,
+    functionName: 'grantRole',
+    args: [
+      UPGRADE_EXECUTOR_ROLE_EXECUTOR, // role
+      account, // account
+    ],
+  });
+
+  // 2. Prepare the transaction (must be called through the UpgradeExecutor)
+  const request = await publicClient.prepareTransactionRequest({
```
TucksonDev, line 45: Only accounts with the ADMIN role can add or remove accounts on the UpgradeExecutor.

The UpgradeExecutor is its own ADMIN, so to make this call, an account with the EXECUTOR role needs to send a call to the UpgradeExecutor, to send a call to its own contract to grant the role.

src/upgradeExecutorFetchPrivilegedAccounts.ts
```diff
@@ -0,0 +1,157 @@
...
+    {
+      indexed: true,
+      internalType: 'address',
+      name: 'sender',
+      type: 'address',
+    },
+  ],
+  name: 'RoleGranted',
+  type: 'event',
+};
```
chrstph-dvx, line 54: If you remove the type on line 31, and add `as const satisfies AbiEvent;`, then Viem is able to infer the type for the args.

src/upgradeExecutorFetchPrivilegedAccounts.ts
```diff
@@ -0,0 +1,157 @@
...
+  });
+  if (!roleRevokedEvents || roleRevokedEvents.length <= 0) {
+    return upgradeExecutorPrivilegedAccounts;
+  }
+
+  // 3. Remove the revoked roles from the result object
+  // (Note: if the same role has been added and revoked multiple times, it will be added and removed multiple times from upgradeExecutorPrivilegedAccounts[account] )
+  roleRevokedEvents.forEach((roleRevokedEvent) => {
+    const account = (roleRevokedEvent.args as RoleRevokedLogArgs).account;
+    const role = (roleRevokedEvent.args as RoleRevokedLogArgs).role;
```
chrstph-dvx, line 143: Not need anymore once you made the change above

src/upgradeExecutorPrepareAddExecutorTransactionRequest.ts
```diff
@@ -0,0 +1,88 @@
...
+  account: Address;
+  upgradeExecutorAddress: Address;
+  executorAccountAddress: Address;
+  publicClient: PublicClient;
+};
+
+/**
+ * Prepares a transaction to grant the executor role to a new account
+ *
+ * - Example: [Add new executor account to UpgradeExecutor](https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/examples/upgrade-executor-add-account/index.ts)
```
chrstph-dvx, line 26: I think we can use Link instead here?

src/upgradeExecutorPrepareRemoveExecutorTransactionRequest.ts
```diff
@@ -0,0 +1,88 @@
...
+ *   publicClient,
+ * });
+ */
+export async function upgradeExecutorPrepareRemoveExecutorTransactionRequest({
+  account,
+  upgradeExecutorAddress,
+  executorAccountAddress,
+  publicClient,
+}: UpgradeExecutorPrepareRemoveExecutorTransactionRequestParams) {
+  const chainId = publicClient.chain!.id;
```
chrstph-dvx, line 50: We can call `validateParentChain` instead here

chrstph-dvx (changes requested):

src/upgradeExecutorPrepareAddExecutorTransactionRequest.ts
TucksonDev: What do you mean? "Add new executor account to UpgradeExecutor" should be a link. I follow the same structure as in `createRollup`.

src/upgradeExecutorFetchPrivilegedAccounts.ts
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * This type is for the result of the upgradeExecutorFetchPrivilegedAccounts function.
+ *
+ * It is an object containing the addresses of the privileged accounts as keys,
+ * and an array with a hash for each role they have.
+ */
+export type UpgradeExecutorPrivilegedAccounts = {
+  // Key: account
+  // Value: array of roles
+  [key: `0x${string}`]: `0x${string}`[];
```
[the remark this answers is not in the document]
TucksonDev: Added an `UpgradeExecutorRole` here and used it as the type for the values: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/feat-add-upgradeexecutorfetchprivilegedaccounts/src/upgradeExecutorEncodeFunctionData.ts#L10

Is that what you mean?

src/upgradeExecutorFetchPrivilegedAccounts.ts
TucksonDev: Removed it, but for some reason now I have to use non-null assertion: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/feat-add-upgradeexecutorfetchprivilegedaccounts/src/upgradeExecutorFetchPrivilegedAccounts.ts#L120

src/upgradeExecutorFetchPrivilegedAccounts.ts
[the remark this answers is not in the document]
chrstph-dvx: Yes, exactly, so we can pass the result to function that accept `UpgradeExecutorRole` without casting

src/upgradeExecutorPrepareRemoveExecutorTransactionRequest.ts
chrstph-dvx: Can we have any case where `publicClient.chain` doesn't exist? If so, I think we should throw error rather than using non-null assertion. Maybe we can create a `validatePublicClient` util

TucksonDev: Following the parentChain example, I added a validateChain helper function: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/b13f0599df92d467c3b1410dbf9adf31870033bd/src/utils/validateChain.ts#L11

chrstph-dvx (approved):
LGTM!

src/upgradeExecutorFetchPrivilegedAccounts.ts
```diff
@@ -0,0 +1,158 @@
...
+  // Value: array of roles
+  [account: `0x${string}`]: UpgradeExecutorRole[];
+};
+
+type RoleGrantedLogArgs = {
+  role: UpgradeExecutorRole;
+  account: `0x${string}`;
+  sender: `0x${string}`;
+};
+type RoleRevokedLogArgs = RoleGrantedLogArgs;
```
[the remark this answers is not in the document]
TucksonDev: Removed both types since we are not using them anymore. Thanks!

PR #30, opened 2023-12-21
feat: add createRollupFetchTransactionHash

### LINKED ISSUE #18: Add `createRollupFetchTransactionHash`
```
function createRollupFetchTransactionHash(rollup: Address)
```

TucksonDev:
This PR adds the function `createRollupFetchTransactionHash` to `src/createRollupFetchTransactionHash.ts`. This function retrieves the transaction hash that created the specified rollup, so it can later be used with `prepareTransaction` or `prepareTransactionReceipt`. It includes a test with a deployed rollup chain. Closes #18 .

Doubts:
- Do we want to include the Arbitrum SDK as a dependency to read the ABIs from there?


src/createRollupFetchTransactionHash.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export async function createRollupFetchTransactionHash(
+  rollupAddress: Address,
+  publicClient: PublicClient,
+) {
+  // Find the RollupInitialized event from that Rollup contract
+  const rollupInitializedEvents = await publicClient.getLogs({
+    address: rollupAddress,
+    event: RollupInitializedEventAbi,
+    fromBlock: 'earliest',
```
spsjvc, line 33: Instead of fetching from `earliest`, maybe we can add the block number for when the rollup creator was deployed at to `wagmi.config.ts` and fetch from then?

src/createRollupFetchTransactionHash.ts
```diff
@@ -0,0 +1,47 @@
...
+      },
+  ],
+  name: "RollupInitialized",
+  type: "event",
+};
+
+export async function createRollupFetchTransactionHash(
+  rollupAddress: Address,
+  publicClient: PublicClient,
+) {
```
spsjvc, line 28: I suggest we always pass objects to these functions, so we have nice parameter naming when we call them

```suggestion
export async function createRollupFetchTransactionHash({
  rollup,
  publicClient
} : {
  rollup: Address,
  publicClient: PublicClient,
}) {
```

src/createRollup.unit.test.ts
spsjvc, line 1: let's rename this to an integration test because it's actually gonna make rpc calls

src/createRollupFetchTransactionHash.ts
```diff
@@ -0,0 +1,47 @@
...
+  // Find the RollupInitialized event from that Rollup contract
+  const rollupInitializedEvents = await publicClient.getLogs({
+    address: rollupAddress,
+    event: RollupInitializedEventAbi,
+    fromBlock: 'earliest',
+    toBlock: 'latest',
+  });
+  if (!rollupInitializedEvents || rollupInitializedEvents.length <= 0) {
+    throw new Error(`No RollupInitialized event found for rollup address ${rollupAddress}`);
+  }
```
spsjvc, line 38: It doesn't seem like `rollupInitializedEvents` can ever be `undefined`, so I think we can just do this:
 
```suggestion
  if (rollupInitializedEvents.length !== 1) {
    throw new Error(`Expected to find 1 RollupInitialized event for rollup address ${rollupAddress} but found ${rollupInitializedEvents.length}`);
  }
```

spsjvc (commented):
looking good! left a few comments

src/createRollupFetchTransactionHash.ts
TucksonDev: Makes sense, but I think I'd have to modify the object then, because each contract in each chain has been deployed at a different block height. Should we create a different PR for that?

src/createRollupFetchTransactionHash.ts
```diff
@@ -0,0 +1,63 @@
...
+      name: 'chainId',
+      type: 'uint256',
+    },
+  ],
+  name: 'RollupInitialized',
+  type: 'event',
+};
+
+export async function createRollupFetchTransactionHash({
+  rollupAddress,
```
spsjvc, line 32: nit: let's just do `rollup` for consistency

spsjvc: let's also export this function at `index.ts`

spsjvc (approved):
Nice, LGTM

PR #35, opened 2024-01-03
feat: keyset calculation script

GreatSoshiant:
In this PR we've added a new API to Orbit SDK to calculate the keyset binary from the desired set of BLS public keys and number of assumed-honest members


spsjvc (approved):
LGTM

PR #37, opened 2024-01-04
feat: add wasmroot verification

Jason-W123:


examples/verify-rollup/src/lib/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+  '0xd1842bfbe047322b3f3b3635b5fe62eb611557784d17ac1d2b1ce9c170af6544', // consensus-v9
+  '0x6b94a7fc388fd8ef3def759297828dc311761e88d8179c7ee8d3887dc554f3c3', // consensus-v10
+  '0xda4e3ad5e7feacb817c21c8d0220da7650fe9051ece68a3f0b1c5d38bbb27b21', // consensus-v10.1
+  '0x0754e09320c381566cc0449904c377a52bd34a6b9404432e80afd573b67f7b17', // consensus-v10.2
+  '0xf559b6d4fa869472dabce70fe1c15221bdda837533dfd891916836975b434dec', // consensus-v10.3
+  '0xf4389b835497a910d7ba3ebfb77aa93da985634f3c052de1290360635be40c4a', // consensus-v11
+];
+
+export const currentMainnetWasmModuleRootIndex = WASMModuleRoots.indexOf(
+  '0x5eF0D09d1E6204141B4d37530808eD19f60FBa35',
```
Jason-W123, line 21: I see we also use the wasmModuleRoot at src/createRollupPrepareConfig.ts, should we create a global constant file for this?

examples/verify-rollup/src/lib/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+  '0x2b20e1490d1b06299b222f3239b0ae07e750d8f3b4dedd19f500a815c1548bbc', // consensus-v7.1
+  '0xd1842bfbe047322b3f3b3635b5fe62eb611557784d17ac1d2b1ce9c170af6544', // consensus-v9
+  '0x6b94a7fc388fd8ef3def759297828dc311761e88d8179c7ee8d3887dc554f3c3', // consensus-v10
+  '0xda4e3ad5e7feacb817c21c8d0220da7650fe9051ece68a3f0b1c5d38bbb27b21', // consensus-v10.1
+  '0x0754e09320c381566cc0449904c377a52bd34a6b9404432e80afd573b67f7b17', // consensus-v10.2
+  '0xf559b6d4fa869472dabce70fe1c15221bdda837533dfd891916836975b434dec', // consensus-v10.3
+  '0xf4389b835497a910d7ba3ebfb77aa93da985634f3c052de1290360635be40c4a', // consensus-v11
+];
+
+export const currentMainnetWasmModuleRootIndex = WASMModuleRoots.indexOf(
```
TucksonDev, line 20: Is this hash correct? I don't see it in WASMModuleRoots, so currentMainnetWasmModuleRootIndex will always be `-1`

examples/verify-rollup/src/partial-handlers/rollup.ts
```diff
@@ -13,10 +13,40 @@ import {
...
+    'wasmModuleRoot',
+  )) as `0x${string}`;
+
+  const index = WASMModuleRoots.indexOf(moduleRoot);
+
+  // check if this wasmModuleRoot belongs to one of mainnet version
+  if (0 <= index) {
+    console.log('The state transition function is standard version');
+    // check if the rollups' arbos version is latest or not
+    if (index < currentMainnetWasmModuleRootIndex) {
```
TucksonDev, line 41: Following with the last comment, this will never be true since I think currentMainnetWasmModuleRootIndex will always be -1.

examples/verify-rollup/src/partial-handlers/rollup.ts
```diff
@@ -13,10 +13,40 @@ import {
...
 const minConfirmPeriodBlocks = BigInt((24 * 60 * 60) / 12.5); // 1 day
 
+export const wasmModuleRootHandler = async (
+  orbitHandler: OrbitHandler,
+  rollupAddress: `0x${string}`,
+): Promise<string> => {
+  let warningMessage: string = '';
+
+  // get the wasmModuleRoot of the given orbit chain
+  const moduleRoot = (await orbitHandler.readContract(
```
TucksonDev, line 28: I think we can add this check inside the `rollup.ts/rollupHandler()` function, instead of creating a new handler for it. What do you think? For now we can run all verifications from verifyRollup, I think.

TucksonDev (changes requested):
Thanks Jason! 🙏  Left a few comments.

examples/verify-rollup/src/lib/constants.ts
Jason-W123: Oh, it is v10 now, sometimes is not -1 is because we just schedule to upgrade but hasn't upgraded: https://github.com/OffchainLabs/nitro/blob/2142d5f4d2e3d967535da8edd0f071803caa0375/Dockerfile#L152

Jason-W123: And the hash is not correct, ty, will change it

examples/verify-rollup/src/partial-handlers/rollup.ts
Jason-W123: I am wondering we can have a separate command to just verify the wasm module root, so I created a new handler here, do you think we don't need to have a separate one? (Because we can also use this to check if they upgrade to our latest arbos version in the future, this is the reason why I created `currentMainnetWasmModuleRootIndex `)

TucksonDev: I don't see a clear reason to have it in a separate command for now. I could be wrong though. If we want to check if a team upgraded to the latest ArbOS version we can run the whole rollup verifier to see if the rest of the parameters are properly set too. In any case, this is not very important now, since there will be changes while we integrate the script in the Orbit SDK, so not 100% sure about this, haha.

Jason-W123: Ok, so I will add it to `rollup.ts/rollupHandler()` first

examples/verify-rollup/src/partial-handlers/rollup.ts
Jason-W123: sometimes is not -1 is because we just schedule to upgrade but hasn't upgraded

examples/verify-rollup/src/partial-handlers/rollup.ts
```diff
@@ -467,5 +468,34 @@ export const rollupHandler = async (
...
+  )) as `0x${string}`;
+
+  const index = WASMModuleRoots.indexOf(moduleRoot);
+
+  // check if this wasmModuleRoot belongs to one of mainnet version
+  if (0 <= index) {
+    console.log('The state transition function is standard version');
+    // check if the rollups' arbos version is latest or not
+    if (index < currentMainnetWasmModuleRootIndex) {
+      warningMessages.push('Arbos version is old');
```
TucksonDev, line 490: Can you add both module roots here? Something like "... Rollup wasmModuleRoot is ${moduleRoot}. Latest standard wasmModuleRoot is ${WASMModuleRoots[currentMainnetWasmModuleRootIndex]}."

TucksonDev (approved):
LGTM! Added a minor comment on the warningMessage to have some extra info, but that's all.
Thanks @Jason-W123 !

PR #38, opened 2024-01-05
feat: generate node config type

spsjvc:
This pull request adds a script that spins up a Nitro container, runs the `--help` command against it, spits the output into a file, parses that file and generates a TypeScript type to be leveraged by the SDK.

It also generates node config for v2.2.5 and makes the SDK use that instead of our hardcoded one.


TucksonDev:
Oh, parsing the help texts must have been fun :D. It looks good, but I have 2 comments:
- We should add a default for --nitro-node-tag in generateNodeConfigType
- I get an error `Syntax error: Bad fd number` when running yarn generate:node-config-type. The reason is that '&>' in `docker run --rm ${nitroNodeImage} --help &> ${nitroNodeHelpOutputFile}` is not supported in all environments. Changing it to `docker run --rm ${nitroNodeImage} --help > ${nitroNodeHelpOutputFile} 2>&1` should work (I think)

TucksonDev (approved):
LGTM 🔥

PR #39, opened 2024-01-13
Chore bump arbos

shotaronowhere:
Arbos 10 includes a couple of parameterization bugs which are resolved in [arbos 11](https://github.com/OffchainLabs/nitro/blob/ce2868328fec5a8b73beec89330d546a8cf88517/arbos/arbosState/arbosstate.go#L304).

- perBatchGasCost is underpricing is fixed. Updated to 240k instead of 100k.
- amortizedCostCapBips bug when set at uint64.max, zero is the correct value to disable the cap

Orbit chains should default to the most recent arbos version.

Also someone left a pk in the .env.example. This is scrubed. You might want to remove it from the git history, but it's too late really.


shotaronowhere:
Whoops, PR already exists #36

PR #44, opened 2024-02-01
feat: add weth gateway registration

TucksonDev:
This PR adds an extra function to set the Wethgateway as the gateway for Weth, since this action is not executed when creating a TokenBridge through the TokenBridgeCreator.

Notable changes:
- A new `createRollupFetchCoreContracts` has also been added which is needed to get the UpgradeExecutor address that will execute this update
- Type `TransactionRequestGasOverrides` has been moved to its own file, since it's now used in two files.
- Integration test has been updated to perform this action


examples/create-token-bridge-eth/index.ts
```diff
@@ -84,6 +86,41 @@ async function main() {
   );
 
   console.log(`Deployed in ${getBlockExplorerUrl(parentChain)}/tx/${txReceipt.transactionHash}`);
+
+  // waiting for retryables to execute, to prevent race conditions
+  await txReceipt.waitForRetryables({ orbitPublicClient: childChainPublicClient });
```
gzeoneth, line 91: this should be sufficient most of the time, but in an edge case where the 2nd retryable is executed (manually) before the 1st, they would both be appeared to executed although the token bridge deployment would fail

we can also check for contract existence

gzeoneth (approved):
LGTM, tho there is a minor edge case

src/createTokenBridge.integration.test.ts
```diff
@@ -151,6 +154,45 @@ it(`successfully deploys token bridge contracts through token bridge creator`, a
   expect(tokenBridgeContracts.orbitChainContracts.beaconProxyFactory).not.toEqual(zeroAddress);
   expect(tokenBridgeContracts.orbitChainContracts.upgradeExecutor).not.toEqual(zeroAddress);
   expect(tokenBridgeContracts.orbitChainContracts.multicall).not.toEqual(zeroAddress);
+
+  // set weth gateway
+  const setWethGatewayTxRequest = await createTokenBridgePrepareSetWethGatewayTransactionRequest({
```
TucksonDev, line 159: We have a problem here. This method will check whether the Weth gateway is already registered. And in the testnode, that's already been fixed, so it will throw an error saying that the gateway is already registered.

We can separate the check for the gateway from the actual prepareRequest. If we do that, we can pass tokenBridgeContracts to the check function. 

It feels a bit weird though. Because PrepareTransactionRequest wouldn't perform the check then, and the check function would receive an object of tokenBridgeContracts instead of a rollup or inbox address (like all other functions do). And again, we would be changing the behavior only because of the test.

I can't find any other solution though.

What do you think @spsjvc ?

TucksonDev: Not applicable anymore since we can now override the TokenBridgeCreator used 💪

src/createTokenBridge.integration.test.ts
```diff
@@ -161,6 +160,46 @@ it(`successfully deploys token bridge contracts through token bridge creator`, a
...
+
+  // checking retryables execution
+  const orbitChainSetGatewayRetryableReceipt = await setWethGatewayTxReceipt.waitForRetryables({
+    orbitPublicClient: nitroTestnodeL2Client,
+  });
+  expect(orbitChainSetGatewayRetryableReceipt).toHaveLength(1);
+  expect(orbitChainSetGatewayRetryableReceipt[0].status).toEqual('success');
+
+  // verify weth gateway
+  const registeredWethGateway = await nitroTestnodeL1Client.readContract({
```
spsjvc, line 196: Instead of doing `nitroTestnodeL1Client.readContract`, let's instead refetch the contracts through the create token bridge tx receipt: `txReceipt.getTokenBridgeContracts`?

PR #45, opened 2024-02-05
feat: update node config for nitro v2.2.x

TucksonDev:
This PR updates `prepareNodeConfig` so it uses the latest node configuration format as of nitro v2.2.2 (including 2.2.3 and 2.2.4)

Most changes are just relocating the configuration since the structure has changed since 2.1.x. However, one more configuration option has been added: `node.delayed-sequencer.use-merge-finality: false,`. This configuration makes the delayed sequencer to not wait for full finality on the parent chain before sequencing messages coming from the parent chain, making it more comfortable to work with cross-chain messages.

As an example, the orbit-setup-script always waits for the cross-chain messages to be executed on the orbit chain. Without this parameter, the orbit-setup-script takes around 40 minutes to fully execute, although it's likely to timeout while waiting for the createTokenBridge retryable execution.

NOTE: Coordination is needed to merge this PR along with the [change in the Orbit setup-script](https://github.com/OffchainLabs/orbit-setup-script/pull/56) and the update of the Orbit SDK package in the Orbit deployment UI.


Tristan-Wilson:
In the block:
`  if (chainConfig.arbitrum.DataAvailabilityCommittee) {`
rpc and rest aggregators should not be the same port. RPC is 9876 and REST is 9877
https://docs.arbitrum.io/node-running/how-tos/data-availability-committee/deploy-a-das#step-3-deploy-the-das

Tristan-Wilson (changes requested):
New stuff looks good, prior comment was on some existing config that should be fixed.

Tristan-Wilson (approved):
LGTM

spsjvc (approved):
LGTM

Tested with 2 chains (on with DAC, one without) on top of Arbitrum Sepolia with v2.2.4:
- https://sepolia.arbiscan.io/tx/0xd664d4cec4ff2ad8bc396e797b76c7c91ec311f48301e74df38d9e3ee42f97ad
- https://sepolia.arbiscan.io/tx/0xa44e14513522cb48dd713eb6f0846a2798884094a30ffe218556965964692993

PR #47, opened 2024-02-08
feat: update contracts

spsjvc:
Adds Ethereum Mainnet and Arbitrum One contracts and updates TokenBridgeCreator addresses to v1.2.0.


TucksonDev (approved):
LGTM.
Not sure about changing the TokenBridge contracts, since we are already doing that in the token-bridge PR, but it's fine either way.

spsjvc:
> LGTM. Not sure about changing the TokenBridge contracts, since we are already doing that in the token-bridge PR, but it's fine either way.

It's ok, they aren't used but good to just do it all in one batch.

PR #53, opened 2024-02-16
feat: reduce default finalize distance

TucksonDev:
This PR adds one more parameter to the node configuration. Since we are using the fast-deposit configuration parameters, we can reduce the number of blocks the chain waits for finalization to 1 (instead of the default 20).


spsjvc (approved):
LGTM

PR #57, opened 2024-02-26
chore: remove redundant create rollup util

spsjvc:
Removes the redundant `createRollup` function. Instead, `createRollupPrepareTransactionRequest` and `createRollupPrepareTransactionReceipt` should be used together to achieve the same result.


src/createRollupTypes.ts
spsjvc, line 1: Moved over from `createRollup.ts` file

TucksonDev: Would it make sense to have it inside the `types` folder?
Or maybe prepareTransactionRequest 🤔

TucksonDev (approved):
LGTM

PR #59, opened 2024-03-01
fix: README instructions

anegg0:
fix: README.
Divided README between SDK consumers and developers 
- Consumers just need to run `yarn install` in their orbit chain repo
- Developers need to add viem and can run integration tests

Fixes TW-183

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 # Arbitrum Orbit SDK
 
 > [!WARNING]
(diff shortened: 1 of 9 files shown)


spsjvc:
@anegg0 the readme covers how you install the package as a consumer, and if you're working on the Orbit SDK, the only thing you need to do is `yarn`

anegg0:
> @anegg0 the readme covers how you install the package as a consumer, and if you're working on the Orbit SDK, the only thing you need to do is `yarn`

Noted, thanks, @spsjvc. Should `yarn add @arbitrum/orbit-sdk viem@^1.20.0` stay in the README, or would you advise replacing it with `yarn`?

PR #60, opened 2024-03-01
fix: Add yarn install to README + minor format fixes

anegg0:
Running `yarn install` is usually obvious, but since a `yarn add` command follows the invocation, it might be nice to mention it.


anegg0:
After cross-checking, this edit is unnecessary.

PR #65, opened 2024-03-15
fix: Ensure networks are registered before SDK calls

chrstph-dvx:
closes fs-361


src/utils/registerNewNetwork.ts
```diff
@@ -0,0 +1,105 @@
...
+    isL2NetworkRegistered = false;
+  }
+
+  if (!isL1NetworkRegistered || !isL2NetworkRegistered) {
+    // register - needed for retryables
+    addCustomNetwork({
+      customL1Network: l1Network,
+      customL2Network: l2Network,
+    });
+  }
```
spsjvc, line 99: There's a case where the L1 (parent) could already be included which would give us an error: https://github.com/OffchainLabs/arbitrum-sdk/blob/main/src/lib/dataEntities/networks.ts#L553

I think this is what we need to do:
```suggestion
  // both networks are already registered, do nothing
  if (isL1NetworkRegistered && isL2NetworkRegistered) { 
    return { l1Network, l2Network };
  }

  if (!isL1NetworkRegistered) {
    // we can assume if parent is not registered, that means the child isn't either, because of this check:
    // https://github.com/OffchainLabs/arbitrum-sdk/blob/main/src/lib/dataEntities/networks.ts#L519
    addCustomNetwork({ customL1Network: l1Network, customL2Network: l2Network });
  } else if (!isL2NetworkRegistered) {
    // parent is already registered, only register the child
    addCustomNetwork({ customL2Network: l2Network });
  }
```

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
```diff
@@ -98,6 +99,12 @@ export async function createTokenBridgePrepareSetWethGatewayTransactionRequest({
 }: CreateTokenBridgePrepareRegisterWethGatewayTransactionRequestParams) {
   const chainId = validateParentChain(parentChainPublicClient);
 
+  // ethers providers
+  const parentChainProvider = publicClientToProvider(parentChainPublicClient);
+  const orbitChainProvider = publicClientToProvider(orbitChainPublicClient);
+
+  await registerNewNetwork(parentChainProvider, orbitChainProvider, rollup);
```
TucksonDev, line 106: Is it possible to move this to inside the `getEstimateForSettingGateway` function? I believe networks only need to be registered when calling the estimation method in the SDK, and it might make sense to do the operation there.

TucksonDev: Also, @spsjvc , what do you think about shifting the responsibility of creating the provider from the publicClient to the createTokenBridge-ethers wrapper? From my point of view, it doesn't really make sense to force prepareTransactionRequest into doing that conversion if it's not really going to use it.

TucksonDev (approved):
LGTM, thanks @chrstph-dvx ! 🚀

spsjvc (approved):
LGTM

PR #67, opened 2024-03-20
feat: Add create token bridge abstraction

chrstph-dvx:
closes FS-388


chrstph-dvx:
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/72/files is fixing the integration tests

src/createTokenBridge.ts
```diff
@@ -0,0 +1,213 @@
...
+      `Transaction hash for retryable is ${orbitChainSetWethGatewayRetryableReceipt[0].transactionHash}`,
+    );
+    if (orbitChainSetWethGatewayRetryableReceipt[0].status !== 'success') {
+      throw new Error(
+        `Retryable status is not success: ${orbitChainSetWethGatewayRetryableReceipt[0].status}. Aborting...`,
+      );
+    }
+  }
+
+  return tokenBridgeContracts;
```
spsjvc, line 212: Let's also return other data, here similar to how we do in [createRollup](https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/createRollup.ts#L198). I think we can add:

- transaction
- transactionReceipt
- retryables (just tx hashes is fine)
- tokenBridgeContracts
- setWethGateway { transaction, transactionReceipt, retryables }

spsjvc: We should make this a type and export it

src/createTokenBridge.ts
```diff
@@ -0,0 +1,213 @@
...
+  rollupOwner,
+  rollupAddress,
+  nativeTokenAddress,
+  parentChainPublicClient,
+  orbitChainPublicClient,
+  createTokenBridgeEnoughCustomFeeTokenAllowanceParamsOverride,
+  createTokenBridgePrepareTransactionRequestParamsOverride,
+  createTokenBridgePrepareRegisterWethGatewayTransactionRequestParamsOverride,
+}: CreateTokenBridgeParams): Promise<TokenBridgeContracts> {
+  if (nativeTokenAddress) {
```
spsjvc, line 90: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/utils/isCustomFeeTokenAddress.ts#L3

src/createTokenBridge.ts
```diff
@@ -0,0 +1,213 @@
...
+ *   }t
+ * });
+ */
+export async function createTokenBridge({
+  rollupOwner,
+  rollupAddress,
+  nativeTokenAddress,
+  parentChainPublicClient,
+  orbitChainPublicClient,
+  createTokenBridgeEnoughCustomFeeTokenAllowanceParamsOverride,
```
spsjvc, line 86: I don't think this is necessary, you can just take the token bridge creator override from `createTokenBridgePrepareTransactionRequestParamsOverride`

spsjvc: I would make the `tokenBridgeCreatorAddressOverride` on the top level of the object and just reuse it everywhere

src/createTokenBridge.ts
```diff
@@ -0,0 +1,213 @@
...
+  createTokenBridgePrepareTransactionRequestParamsOverride,
+  createTokenBridgePrepareRegisterWethGatewayTransactionRequestParamsOverride,
```
spsjvc, line 88:
```suggestion
  setWethGatewayOverrides,
```

src/createTokenBridge.ts
```diff
@@ -0,0 +1,213 @@
...
+ * });
+ */
+export async function createTokenBridge({
+  rollupOwner,
+  rollupAddress,
+  nativeTokenAddress,
+  parentChainPublicClient,
+  orbitChainPublicClient,
+  createTokenBridgeEnoughCustomFeeTokenAllowanceParamsOverride,
+  createTokenBridgePrepareTransactionRequestParamsOverride,
```
spsjvc, line 87: Again, I think just moving out `gasOverrides` and `retryableOverrides` to top level is better here

src/createTokenBridge.ts
```diff
@@ -0,0 +1,298 @@
...
+   */
+  transaction: Address;
+  /**
+   * Transaction receipt of createTokenBridgePrepareTransactionReceipt ({@link CreateTokenBridgeTransactionReceipt})
+   */
+  transactionReceipt: CreateTokenBridgeTransactionReceipt;
+  /**
+   * Retryable transaction receipts of createTokenBridgePrepareTransactionReceipt ({@link WaitForRetryablesResult})
+   */
+  retryables: WaitForRetryablesResult;
```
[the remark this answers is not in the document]
chrstph-dvx: `WaitForRetryablesResult` resolve to `[TransactionReceipt, TransactionReceipt]`. If we ever update `WaitForRetryablesResult` to something else, then it will be updated here too

src/createTokenBridge.ts
```diff
@@ -0,0 +1,298 @@
...
+     */
+    transaction: Address;
+    /**
+     * Transaction receipt of createTokenBridgePrepareSetWethGatewayTransactionReceipt ({@link createTokenBridgePrepareSetWethGatewayTransactionReceipt})
+     */
+    transactionReceipt: CreateTokenBridgeSetWethGatewayTransactionReceipt;
+    /**
+     * Retryable transaction receipts of createTokenBridgePrepareSetWethGatewayTransactionReceipt ({@link WaitForRetryablesResult})
+     */
+    retryables: TransactionReceipt[];
```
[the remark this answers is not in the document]
chrstph-dvx: If we only have one, we might as well return the item directly, rather than an array no?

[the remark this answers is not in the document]
chrstph-dvx: discussed offline, we're returning an array to be consistent with other methods

spsjvc (approved):
LGTM

PR #68, opened 2024-03-21
feat: add high level createRollup function

fionnachan:
- [x] add `createRollup` function
- [x] rename `createRollupParams` to `createRollupTxParams`
- [x] replace `createTestRollup` usage with `createRollup` in [src/createRollup.integration.test.ts](https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/68/files#diff-3a800720e92f7991b6052532ba68e97fe7830e16defb7be19de9b21768ac3c38)
- [x] delete `testHelper.ts`
- [x] update examples
- [x] add custom gas token create rollup integration test

closes FS-387
closes FS-320


src/index.ts
```diff
@@ -12,7 +12,7 @@ import {
   createRollupPrepareCustomFeeTokenApprovalTransactionRequest,
   CreateRollupPrepareCustomFeeTokenApprovalTransactionRequestParams,
 } from './createRollupPrepareCustomFeeTokenApprovalTransactionRequest';
-import { CreateRollupFunctionInputs, CreateRollupParams } from './types/createRollupTypes';
+import { CreateRollupFunctionInputs, CreateRollupTxParams } from './types/createRollupTypes';
```
chrstph-dvx, line 15: I get the point of the changes, but because it's exported to consumers, we might not want to change it? Or maybe rename it before we export?

It's just a type, so it might be overkill

src/createRollup.ts
```diff
@@ -0,0 +1,202 @@
...
+ * } = await createRollup({
+ *   config: createRollupConfig,
+ *   rollupOwner: deployer,
+ *   batchPoster,
+ *   validators,
+ *   parentChainPublicClient,
+ * });
+ */
+export async function createRollup({
+  rollupOwner,
```
spsjvc, line 150:
```suggestion
  account
```

src/createRollup.ts
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * const {
+ *   transaction,
+ *   transactionReceipt,
+ *   coreContracts,
+ * } = await createRollup({
+ *   config: createRollupConfig,
+ *   rollupOwner: deployer,
+ *   batchPoster,
+ *   validators,
```
spsjvc, line 145: I'm thinking we group all of the params that are actual args to the contract into one `params` field or something like that

src/createRollup.integration.test.ts
```diff
@@ -1,56 +1,121 @@
...
-    // assert the core contracts were successfully obtained
-    expect(createRollupInformation.coreContracts).toBeDefined();
+  // create test rollup with ETH as gas token
+  return {
+    createRollupConfig,
+    createRollupInformation,
+  };
+}
+
+describe(`createRollup`, () => {
```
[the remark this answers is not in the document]
fionnachan: ok, this is from the existing code, i'll delete it

src/createRollup.integration.test.ts
```diff
@@ -1,56 +1,121 @@
...
-describe(`createRollup`, async () => {
-  // create test rollup
-  const createRollupInformation = await createTestRollup({
-    deployer,
+async function createRollupHelper(nativeToken: Address = zeroAddress) {
+  const chainId = generateChainId();
+
+  const createRollupConfig = createRollupPrepareConfig({
+    chainId: BigInt(chainId),
+    owner: userTokenBridgeDeployer.address,
```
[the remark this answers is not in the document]
fionnachan: we need to add it because it was the only account that received the custom token deployed by the testnode script from the `--l3-fee-token` flag

it's named `userTokenBridgeDeployer` because it's the `user_token_bridge_deployer` from nitro testnode

src/createRollup.ts
fionnachan: do you mean something like this?
```typescript
createRollup(createRollupParams, { account, parentChainPublicClient })
```

spsjvc: Same as here https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/createRollupPrepareTransactionRequest.ts#L32

src/createRollup.ts
```diff
@@ -0,0 +1,210 @@
...
+import { getBlockExplorerUrl } from './utils/getters';
+import {
+  CreateRollupTransaction,
+  createRollupPrepareTransaction,
+} from './createRollupPrepareTransaction';
+import { CreateRollupParams } from './types/createRollupTypes';
+
+type PublicClientWithDefinedChain = Omit<PublicClient, 'chain'> & {
+  chain: Chain;
+};
```
spsjvc, line 20: this could be useful elsewhere, so let's move it to `src/types/ParentChain.ts`
we can add this:

```

export type ParentChainPublicClient = Prettify<
  Omit<PublicClient, 'chain'> & {
    chain: Prettify<Omit<Chain, 'id'> & { id: ParentChainId }>;
  }
>;

export function validateParentChainPublicClient(
  publicClient: PublicClient,
): ParentChainPublicClient {
  const chainId = publicClient.chain?.id;

  if (!isValidParentChainId(chainId)) {
    throw new Error(`Parent chain not supported: ${chainId}`);
  }

  return publicClient as ParentChainPublicClient;
}
```

and later use it as:

```
const validatedPublicClient = validateParentChainPublicClient(publicClient)
```

spsjvc (approved):
LGTM

PR #72, opened 2024-04-05
fix: Fix registerNewNetworks for l2 networks

chrstph-dvx:
Fix an issue when we would call `addCustomNetworks({ customL1Network: arbitrumChain, customL2Network: orbitChain })`, which registered the `arbitrumChain` as a L1


src/utils/registerNewNetwork.ts
```diff
@@ -74,38 +74,75 @@ export const registerNewNetwork = async (
...
+        outbox: await rollup.outbox(),
+        rollup: rollup.address,
+        sequencerInbox: await rollup.sequencerInbox(),
+      },
+      explorerUrl: '',
+      isArbitrum: true,
+      isCustom: true,
+      partnerChainIDs: [childNetworkInfo.chainId],
+      retryableLifetimeSeconds: 7 * 24 * 60 * 60,
+      partnerChainID: 1,
```
chrstph-dvx, line 96: We don't have any information about the L1 parent of the L2 chain here, default to Mainnet

src/utils/registerNewNetwork.ts
```diff
@@ -74,38 +75,75 @@ export const registerNewNetwork = async (
...
-  let isL1NetworkRegistered = false;
-  try {
-    await getL1Network(l1Provider);
-    isL1NetworkRegistered = true;
-  } catch {
-    isL1NetworkRegistered = false;
-  }
+  // If the parent chain is an Arbitrum chain, we need to register parent and child chain both as L2 chains
+  if (await isArbitrumChain(parentProvider)) {
+    const isParentNetworkRegistered = await isNetworkRegistered(parentProvider, { type: 'L2' });
```
spsjvc, line 80: I'd maybe move the construction of these `L1Network` and `L2Network` objects outside of this function so it's cleaner to read

src/utils/registerNewNetwork.integration.test.ts
```diff
@@ -0,0 +1,72 @@
...
+});
+
+const nitroTestnodeL3Client = createPublicClient({
+  chain: nitroTestnodeL3,
+  transport: http(nitroTestnodeL3.rpcUrls.default.http[0]),
+});
+
+const l1Client = publicClientToProvider(nitroTestnodeL1Client);
+const l2Client = publicClientToProvider(nitroTestnodeL2Client);
+const l3Client = publicClientToProvider(nitroTestnodeL3Client);
```
spsjvc, line 27:
```suggestion
const l1Provider = publicClientToProvider(nitroTestnodeL1Client);
const l2Provider = publicClientToProvider(nitroTestnodeL2Client);
const l3Provider = publicClientToProvider(nitroTestnodeL3Client);
```

src/utils/registerNewNetwork.ts
```diff
@@ -5,38 +5,39 @@ import {
...
 export const registerNewNetwork = async (
-  l1Provider: JsonRpcProvider,
-  l2Provider: JsonRpcProvider,
+  parentProvider: JsonRpcProvider,
+  childProvider: JsonRpcProvider,
   rollupAddress: string,
 ): Promise<{
-  l1Network: L1Network;
+  l1Network: L1Network | Omit<L2Network, 'tokenBridge'>;
   l2Network: Omit<L2Network, 'tokenBridge'>;
```
spsjvc, line 33: Maybe return `parentNetwork` and `childNetwork` here

src/utils/registerNewNetwork.integration.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+it('should register L1 and L2 respectively as L1 and L2', async () => {
+  const testnodeInformation = getInformationFromTestnode();
+  const { l1Network, l2Network } = await registerNewNetwork(
+    l1Client,
+    l2Client,
+    testnodeInformation.rollup,
+  );
+
+  expect(l1Network).toHaveProperty('chainID', nitroTestnodeL1.id);
```
spsjvc, line 37: this seems simpler
```suggestion
  expect(l1Network.chainID).toEqual(nitroTestnodeL1.id);
```

src/utils/registerNewNetwork.integration.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  expect(l2Network).toHaveProperty('chainID', nitroTestnodeL2.id);
+  expect(Object.keys(l1Networks)).toContain(nitroTestnodeL1.id.toString());
+  expect(Object.keys(l2Networks)).toContain(nitroTestnodeL2.id.toString());
+  expect(l1Network.partnerChainIDs).toContain(nitroTestnodeL2.id);
+  expect(l2Network.partnerChainID).toBe(l1Network.chainID);
```
spsjvc, line 42: I would stay away from using `toBe` and use `toEqual`. I know there's no difference for primitive types, but don't want to start a trend

src/utils/registerNewNetwork.ts
```diff
@@ -1,42 +1,44 @@
...
-  const l2Network: L2Network = {
-    chainID: l2NetworkInfo.chainId,
+async function createChildNetwork({
+  rollupAddress,
+  provider,
+  parentNetwork,
+  childNetwork,
+}: {
+  rollupAddress: string;
+  provider: JsonRpcProvider;
```
spsjvc, line 35: should be explicit here
```suggestion
  parentProvider: JsonRpcProvider;
```

src/utils/registerNewNetwork.ts
```diff
@@ -73,39 +75,144 @@ export const registerNewNetwork = async (
...
+}: {
+  parentNetwork: Network;
+  childNetwork: Network;
+  isArbitrum: TIsArbitrum;
+}): Promise<L1Network | L2Network> {
+  if (isArbitrum) {
+    return {
+      chainID: parentNetwork.chainId,
+      name: parentNetwork.name,
+      confirmPeriodBlocks: 0,
```
spsjvc, line 102: my bad here, even though this should never be used let's still put in a reasonable value
```suggestion
      confirmPeriodBlocks: 45818,
```

src/utils/registerNewNetwork.ts
```diff
@@ -73,39 +75,144 @@ export const registerNewNetwork = async (
...
+      isArbitrum: true,
+      isCustom: true,
+      partnerChainIDs: [childNetwork.chainId],
+      retryableLifetimeSeconds: 7 * 24 * 60 * 60,
+      partnerChainID: sepolia.id,
+      nitroGenesisBlock: 0,
+      nitroGenesisL1Block: 0,
+      depositTimeout: 1800000,
+      blockTime: arbitrumSdkConstants.ARB_MINIMUM_BLOCK_TIME_IN_SECONDS,
+    } as L2Network;
```
spsjvc, line 136: don't think this cast is necessary?

src/utils/registerNewNetwork.ts
```diff
@@ -73,39 +75,144 @@ export const registerNewNetwork = async (
...
-    isL2NetworkRegistered = false;
+  return {
+    blockTime: 10,
+    chainID: parentNetwork.chainId,
+    explorerUrl: '',
+    isCustom: true,
+    name: parentNetwork.name,
+    partnerChainIDs: [childNetwork.chainId],
+    isArbitrum: false,
+  } as L1Network;
```
spsjvc, line 147: don't think this cast is necessary?

src/utils/registerNewNetwork.ts
```diff
@@ -1,42 +1,54 @@
...
-    isArbitrum: false,
-  };
+const parentChainIdFromChildChainId = (childChainId: number) => {
+  return (
+    {
+      [arbitrum.id]: mainnet.id,
+      [arbitrumNova.id]: mainnet.id,
+      [arbitrumSepolia.id]: sepolia.id,
+    }[childChainId] || sepolia.id
+  );
```
spsjvc, line 35: I think we also want the nitro testnode L2 here, and maybe not default sepolia and just let it error?

spsjvc (approved):
LGTM

PR #73, opened 2024-04-08
fix: set max data size for all parent chains

spsjvc:


src/types/ParentChain.ts
```diff
@@ -1,7 +1,12 @@
 import { PublicClient } from 'viem';
-import { chains } from '../chains';
 
-export type ParentChain = (typeof chains)[number];
+import { chains, nitroTestnodeL3 } from '../chains';
+
+export type ParentChain = Exclude<
+  (typeof chains)[number],
+  // exclude nitro-testnode L3 from the list of parent chains
+  { id: typeof nitroTestnodeL3.id }
```
spsjvc, line 8: filtering out nitro testnode L3 as we don't want to support it as parent chain for L4s

TucksonDev (approved):
LGTM
(Pending integration tests execution)

PR #74, opened 2024-04-09
feat: add get contracts script (orbit verifier)

TucksonDev:
This PR adds a new script to get all contracts of a chain (both rollup and tokenbridge)
This PR also updates some of the outputs to make them clearer


examples/verify-rollup/src/partial-handlers/rollup.ts
```diff
@@ -485,17 +486,17 @@ export const rollupHandler = async (
 
   // check if this wasmModuleRoot belongs to one of mainnet version
   if (0 <= index) {
-    console.log('The state transition function is standard version');
+    console.log(`WASM module root recognized: ${moduleRoot}`);
     // check if the rollups' arbos version is latest or not
-    if (index < latestWasmModuleRootIndex) {
+    /* if (index < latestWasmModuleRootIndex) {
```
TucksonDev, line 491: This is commented out to handle different wasm module roots in a better way (I believe it shouldn't be a warning). There's a new task for this [here](https://linear.app/offchain-labs/issue/INT-199/verify-the-wasm-module-root-against-the-arbos-version).

PR #76, opened 2024-04-12
fix: smol rollup fixes for L2 chains

TucksonDev:
When using the SDK to deploy an L2 orbit chain, some of the default examples will fail to execute. This PR adds fixes for those cases:
- prepareNodeConfig should receive a beacon RPC URL to add to the node configuration (even if it's not used at the beginning, it is safer to have the beacon RPC in the config since now, to prevent potential issues when changing to posting blobs)
- createTokenBridge fails to find the rollup contracts in some L1s due to a bug that searches logs from 'earliest' instead of the block where the RollupCreator was deployed.


src/createRollupFetchTransactionHash.ts
```diff
@@ -59,7 +59,7 @@ export async function createRollupFetchTransactionHash({
   const chainId = validateParentChain(publicClient);
 
   const fromBlock =
-    chainId in Object.keys(earliestRollupCreatorDeploymentBlockNumber)
+    chainId in earliestRollupCreatorDeploymentBlockNumber
```
TucksonDev, line 62: Although we are now searching from the block where the RollupCreator was deployed, it's still too many blocks to search through for some free tier L1 nodes. We might want to find a way to adjust the search limits and perform multiple searches.

src/utils/getParentChainLayer.ts
```diff
@@ -0,0 +1,34 @@
...
+    // L3 mainnet
+    case arbitrumOne.id:
+    case arbitrumNova.id:
+    // L3 testnet
+    case arbitrumSepolia.id:
+    // L3 nitro-testnode
+    case nitroTestnodeL2.id:
+      return 2;
+  }
+}
```
TucksonDev, line 34: Not sure if this is the best place for this file, @spsjvc .
Also, I used simply "1" and "2" for the layers, should I use some constants there?

src/createRollup.ts
```diff
@@ -180,16 +180,19 @@ export async function createRollup({
...
     await validatedParentChainPublicClient.waitForTransactionReceipt({ hash: txHash }),
   );
 
+  // get the transaction
+  // (we get the transaction after waiting for the receipt, to prevent `TransactionNotFoundError`s coming
+  // from RPCs that use load balancing and caching. More information can be found here:
+  // https://github.com/wevm/viem/issues/1056#issuecomment-1689800265 )
+  const tx = createRollupPrepareTransaction(
+    await validatedParentChainPublicClient.getTransaction({ hash: txHash }),
+  );
```
TucksonDev, line 194: This is needed to prevent cases where the RPC fails to find the transaction that was just sent (due to load balancing or caching). For example, when using the public ArbSepolia RPC, calling this method fails when requesting the transaction. More information here: https://github.com/wevm/viem/issues/1056#issuecomment-1689800265

src/prepareNodeConfig.ts
```diff
@@ -138,6 +142,12 @@ export function prepareNodeConfig({
     },
   };
 
+  if (parentChainBeaconRpcUrl) {
```
spsjvc, line 145: maybe we should add runtime validation that this must be provided if the parent chain is an L1, not just in the example

src/utils/getParentChainLayer.ts
spsjvc: I think it's good

src/prepareNodeConfig.ts
TucksonDev: Added here: https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/76/commits/f6e550ee64b58b7d8cda537e25f1a5d76eea956d

spsjvc (approved):
LGTM

PR #80, opened 2024-04-18
feat: add AEP fee routing utils with examples

TucksonDev:
This PR adds the functionality needed to provide an example script that an L2 Orbit chain can use to setup the appropriate fee routing to comply with the Arbitrum Expansion Program requirement of sharing 10% of the revenue back to the Arbitrum ecosystem ([source](https://forum.arbitrum.foundation/t/the-arbitrum-expansion-program-and-developer-guild/20722)).

Flow for the example script:
1. Obtain all chain contracts (needed to perform the operations)
2. Obtain the current fee collector for the active batch poster of the chain
3. Deploy the ChildToParentRouter contract, configured to send the amounts received to the Arbitrum Foundation L1 multisig address
4. Deploy the RewardDistributor contract, configured to distribute 90% of the amounts received to the current configured fee collector, and 10% to the ChildToParentRouter contract
5. Set the fee collector of the batch poster to the RewardDistributor contract

The example script lives in `examples/setup-aep-fee-routing`.

An additional example script has been provided in `examples/setup-fee-distributor-contract` to show how to set the RewardDistributor for one fee type, with a link to additional instructions on how to set it up for other types.

Functionality added:
- `feeRouterDeployChildToParentRouter` and `feeRouterDeployRewardDistributor` helper functions to easily deploy these contracts
- ~`ArbAggregatorActions` decorator to be able to interact with this precompile~ (Removed to be added in a different PR)

ToDo:
- [x] Add unit tests for the deploy helper functions
- [x] Add integration tests for the ArbAggregatorActions
- [x] ~Confirm that the bytecode of the contracts to be deployed is correct~ (we are now using the npm released package)
- [x] ~Provide instructions on how to obtain this bytecode from the source code~ (shouldn't be necessary I think)
- [x] Write a README for the example script
- [x] Add an example script to only deploy a RewardDistributor, and provide instructions on how to set those for each fee type, linking to docs


examples/setup-aep-fee-router/index.ts
```diff
@@ -0,0 +1,283 @@
...
+
+if (typeof process.env.ORBIT_CHAIN_RPC === 'undefined') {
+  throw new Error(`Please provide the "ORBIT_CHAIN_RPC" environment variable`);
+}
+
+// load the chain owner account (or any account that has the executor role in the UpgradeExecutor)
+const chainOwner = privateKeyToAccount(sanitizePrivateKey(process.env.CHAIN_OWNER_PRIVATE_KEY));
+
+// set the parent chain and create a public client for it
+const parentChain = arbitrumSepolia;
```
DZGoldman, line 44: shouldn't this be provided as an input?

DZGoldman: below the script appears to assume the parent chain is mainnet (hardcoded 0x1Afa41C006dA1605846271E7bdae942F2787f941 musig address)

src/feeRouterDeployChildToParentRouter.ts
```diff
@@ -0,0 +1,142 @@
...
+    orbitChainWalletClient: WalletClient;
+    parentChainTargetAddress: Address;
+    minDistributionInvervalSeconds?: bigint;
+    rollup?: Address;
+    parentChainTokenAddress?: Address;
+  }>
+>;
+
+// Default minimum distribution interval seconds
+const DEFAULT_MIN_DISTRIBUTION_INVERVAL_SECONDS = BigInt(60 * 60 * 24); // 1 day
```
DZGoldman, line 24: I suggest a default of 1 week instead (this is what we're using for Nova fee routing). The griefing vector is one outbox execution per interval (note that these fees will be paid by the AF/DAO, not the orbit chain owner)

examples/setup-fee-distributor-contract/index.ts
```diff
@@ -0,0 +1,175 @@
...
+
+if (typeof process.env.ORBIT_CHAIN_RPC === 'undefined') {
+  throw new Error(`Please provide the "ORBIT_CHAIN_RPC" environment variable`);
+}
+
+// load the chain owner account (or any account that has the executor role in the UpgradeExecutor)
+const chainOwner = privateKeyToAccount(sanitizePrivateKey(process.env.CHAIN_OWNER_PRIVATE_KEY));
+
+// creating some random addresses to show the example
+const randomAccount = privateKeyToAccount(generatePrivateKey());
```
DZGoldman, line 42: I guess I'm unclear on wether or not these scripts are intended to be run against mainnet (if so, these should be provided as inputs, maybe as a json file)

DZGoldman (changes requested):

src/feeRouterDeployChildToParentRouter.ts
TucksonDev: Oh, right. I forgot to ask about this. Changed. Thanks!

examples/setup-aep-fee-router/index.ts
TucksonDev: I've added the parent chain target address as an env variable. All these scripts are meant to be only examples that one can copy and adjust as necessary, and they follow the same structure: parent chain arbSepolia and inputs as env variables or random information (like the chain id, batch poster and staker wallet when creating a rollup for example). Happy to discuss further though

examples/setup-fee-distributor-contract/index.ts
TucksonDev: These scripts are meant to be examples only. They can be run on mainnet, but should be configured first (even if you run it on testnet, you'll have to configure a few things). Random data is used sometimes like for the batch poster in the createRollup example.

TucksonDev: I tried to provide a description of how to configure the different recipient addresses and their weights though.

examples/setup-aep-fee-router/index.ts
TucksonDev: Examples were modified so all inputs are now defined in the .env file.

examples/setup-fee-distributor-contract/index.ts
TucksonDev: Examples were modified so all inputs are now provided in the .env file

spsjvc (approved):

PR #87, opened 2024-05-01
feat: script for setting l2 base fee

renlulu:


spsjvc:
@renlulu thanks for the contribution!

This operation should be done through the ArbOwner actions, you can have a look at an example [here](https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/decorators/arbOwnerPublicActionsUpgradeExecutor.integration.test.ts#L113).

We can keep this PR open if you want to update to this approach.

renlulu:
Okie, I will check out

PR #88, opened 2024-05-06
chore: Bump viem to v2

chrstph-dvx:
Closes FS-279


src/arbOwnerPrepareTransactionRequest.ts
```diff
@@ -5,45 +5,57 @@ import {
...
+  abi,
+  args,
+}: ArbOwnerEncodeFunctionDataParameters<TFunctionName>): EncodeFunctionDataReturnType {
   return encodeFunctionData({
     abi,
     functionName,
     args,
-  });
+  } as EncodeFunctionDataParameters);
 }
```
chrstph-dvx, line 39: Applies to all `xxxEncodeFunctionData`: we could probably remove them and directly call `encodeFunctionData(params)` in `xxxprepareFunctionData`. It would simplify type

PR #91, opened 2024-05-07
feat: add ArbAggregator decorator

TucksonDev:
This PR adds a decorator to use the ArbAggregator precompile

### DIFF
--- a/src/arbAggregatorPrepareTransactionRequest.ts
+++ b/src/arbAggregatorPrepareTransactionRequest.ts
@@ -9,34 +9,41 @@ import {
 
 import { arbAggregator } from './contracts';
 import { upgradeExecutorEncodeFunctionData } from './upgradeExecutor';
-import { Prettify } from './types/utils';
+import { GetFunctionName } from './types/utils';
 
-type ArbAggregatorEncodeFunctionDataParameters = Prettify<
-  Omit<EncodeFunctionDataParameters<typeof arbAggregator.abi, string>, 'abi'>
->;
+type ArbAggregatorAbi = typeof arbAggregator.abi;
+export type ArbAggregatorPrepareTransactionRequestFunctionName = GetFunctionName<ArbAggregatorAbi>;
+export type ArbAggregatorEncodeFunctionDataParameters<
+  TFunctionName extends ArbAggregatorPrepareTransactionRequestFunctionName,
+> = EncodeFunctionDataParameters<ArbAggregatorAbi, TFunctionName>;
 
-function arbAggregatorEncodeFunctionData({
-  functionName,
-  args,
-}: ArbAggregatorEncodeFunctionDataParameters) {
+function arbAggregatorEncodeFunctionData<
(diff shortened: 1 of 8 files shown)


src/arbAggregatorPrepareTransactionRequest.ts
```diff
@@ -0,0 +1,84 @@
...
+  Transport,
+} from 'viem';
+
+import { arbAggregator } from './contracts';
+import { upgradeExecutorEncodeFunctionData } from './upgradeExecutor';
+import { Prettify } from './types/utils';
+
+type ArbAggregatorEncodeFunctionDataParameters = Prettify<
+  Omit<EncodeFunctionDataParameters<typeof arbAggregator.abi, string>, 'abi'>
+>;
```
chrstph-dvx, line 16: using `string` here means we won't infer `args` based on the function name. You can have a look at `arbOwnerPrepareTransactionRequest`

TucksonDev: Thanks Chris!
It looks like I copied the previous code when creating the decorator, and it definitely looks nicer now. I've made the necessary changes (waiting for the tests to finish)

src/decorators/arbAggregatorActions.integration.test.ts
```diff
@@ -0,0 +1,60 @@
...
+        account: l2RollupOwner.address,
+      });
+    await nitroTestnodeL2Client.sendRawTransaction({
+      serializedTransaction: await l2RollupOwner.signTransaction(setFeeCollectorTransactionRequest),
+    });
+
+    // Check the fee collector has changed
+    const batchPosterFeeCollector = await nitroTestnodeL2Client.arbAggregatorReadContract({
+      functionName: 'getFeeCollector',
+      args: [batchPosters[1] as `0x${string}`],
```
chrstph-dvx, line 56: NIT: this should not be needed, type should be infered.

src/decorators/arbAggregatorActions.integration.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    const batchPosters = await nitroTestnodeL2Client.arbAggregatorReadContract({
+      functionName: 'getBatchPosters',
+    });
+
+    expect(batchPosters).toHaveLength(2);
+    expect(batchPosters[0]).toEqual('0xA4b000000000000000000073657175656e636572');
+
+    const batchPosterFeeCollector = await nitroTestnodeL2Client.arbAggregatorReadContract({
+      functionName: 'getFeeCollector',
+      args: [batchPosters[0] as `0x${string}`],
```
chrstph-dvx, line 29: This should not be needed now, return type should be infered (same below on L45 and 56)

chrstph-dvx (commented):
Overall LGTM, just one minor comment.

chrstph-dvx (approved):

chrstph-dvx (approved):

PR #92, opened 2024-05-07
ci: cancel in progress workflows

chrstph-dvx:
New workflow will cancel previous workflow for a given PR

closes fs-481


chrstph-dvx:
https://github.com/OffchainLabs/arbitrum-orbit-sdk/actions/runs/8991708126 was canceled as title was edited

fionnachan (approved):

PR #100, opened 2024-05-23
feat: create rollup deployment params config

spsjvc:
This function replaces `createRollupPrepareConfig` (which is now marked as deprecated), and it preserves the same signature apart from the requirement of the parent public client, so we can make config changes based on the parent chain. It replaces all usage across the codebase. The old function is still there for backwards compat.

The reason why it's named `createRollupPrepareDeploymentParamsConfig`: https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/generated.ts#L792

Closes FS-545.

### DIFF
--- a/src/createRollupPrepareDeploymentParamsConfig.ts
+++ b/src/createRollupPrepareDeploymentParamsConfig.ts
@@ -1,0 +1,43 @@
+import { PublicClient } from 'viem';
+
+import { ChainConfig } from './types/ChainConfig';
+import { CreateRollupFunctionInputs } from './types/createRollupTypes';
+import { prepareChainConfig } from './prepareChainConfig';
+import { validateParentChain } from './types/ParentChain';
+import { defaults } from './createRollupPrepareDeploymentParamsConfigDefaults';
+import { getDefaultConfirmPeriodBlocks } from './getDefaultConfirmPeriodBlocks';
+import { getDefaultSequencerInboxMaxTimeVariation } from './getDefaultSequencerInboxMaxTimeVariation';
+
+type RequiredKeys = 'chainId' | 'owner';
+
+export type CreateRollupPrepareConfigResult = CreateRollupFunctionInputs[0]['config'];
+
+export type CreateRollupPrepareConfigParams = Pick<CreateRollupPrepareConfigResult, RequiredKeys> &
+  Partial<Omit<CreateRollupPrepareConfigResult | 'chainConfig', RequiredKeys>> & {
+    chainConfig?: ChainConfig;
+  };
+
+export function createRollupPrepareDeploymentParamsConfig(
+  client: PublicClient,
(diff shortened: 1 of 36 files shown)


src/getDefaultConfirmPeriodBlocks.ts
```diff
@@ -0,0 +1,12 @@
...
+import { parentChainIsMainnet } from './parentChainIsMainnet';
+
+export function getDefaultConfirmPeriodBlocks(parentChainId: ParentChainId): bigint {
+  const isMainnet = parentChainIsMainnet(parentChainId);
+
+  if (!isMainnet) {
+    return 150n;
+  }
+
+  return 45_818n;
```
spsjvc, line 11: for base, just x6 this

src/getDefaultSequencerInboxMaxTimeVariation.ts
```diff
@@ -0,0 +1,19 @@
...
+  futureBlocks: bigint;
+  delaySeconds: bigint;
+  futureSeconds: bigint;
+};
+
+export function getDefaultSequencerInboxMaxTimeVariation(
+  parentChainId: ParentChainId,
+): SequencerInboxMaxTimeVariation {
+  return {
+    delayBlocks: BigInt(5_760),
```
spsjvc, line 14: 4 days

src/getDefaultSequencerInboxMaxTimeVariation.ts
```diff
@@ -0,0 +1,19 @@
...
+    futureBlocks: BigInt(48),
+    delaySeconds: BigInt(86_400),
```
spsjvc, line 16: 4 days in seconds, blocks just divide by 12

spsjvc: for base, multiply blocks by 6

src/createRollupPrepareDeploymentParamsConfig.unit.test.ts
spsjvc, line 1: These are the same test cases used in `createRollupPrepareConfig.unit.test.ts`

src/createRollupPrepareDeploymentParamsConfig.ts
```diff
@@ -0,0 +1,50 @@
...
+  RequiredParams & { chainConfig?: ChainConfig } & OptionalParams
+>;
+
+export function createRollupPrepareDeploymentParamsConfig(
+  client: PublicClient,
+  { chainConfig, ...params }: CreateRollupPrepareDeploymentParamsConfigParams,
+): CreateRollupPrepareDeploymentParamsConfigResult {
+  const parentChainId = validateParentChain(client);
+
+  const defaultsBasedOnParentChain = {
```
spsjvc, line 33: This is what's different with the new implementation

chrstph-dvx (approved):

PR #102, opened 2024-05-28
feat: add `getValidators`

chrstph-dvx:
Add `getValidators` util to retrieve all validators for a given rollup

closes fs-519, fs-556


src/getValidators.ts
```diff
@@ -0,0 +1,120 @@
...
+        );
+        return [];
+      }
+      default: {
+        console.warn(`[getValidators] unknown 4bytes, tx id: ${tx.hash}`);
+      }
+    }
+  });
+
+  return validators.filter((validator) => validator) as `0x${string}`[];
```
fionnachan, line 119: we need to call `isValidator` somewhere to filter out the addresses which are no longer a validator (e.g. which was called with `setValidator([address],[false])`)

src/getValidators.unit.test.ts
fionnachan, line 1: i'd add one or two more tests for chains with >1 validator

fionnachan (commented):
we should write some test cases with validators being changed using `setValidator` on the testnode L2 for integration tests

src/getValidators.ts
```diff
@@ -0,0 +1,132 @@
...
+          abi: rollupAdminLogicABI,
+          data: tx.input,
+        });
+
+        if (typeof decodedValidator === 'string' && typeof isAdd === 'boolean') {
+          if (isAdd) {
+            acc.add(decodedValidator);
+          } else {
+            acc.delete(decodedValidator);
+          }
```
fionnachan, line 85: perfect! haha

src/decorators/getValidators.integration.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    serializedTransaction: await l3RollupOwner.signTransaction(tx),
+  });
+}
+
+describe('successfully get validators', async () => {
+  it('when disabling the same validator multiple time', async () => {
+    const randomAccount = privateKeyToAccount(generatePrivateKey()).address;
+
+    const initialValidators = await getValidators(client, { rollupAddress: l3Rollup });
+    // By default, nitro chains have 10 validators
```
fionnachan, line 41: Where does this default come from?

Very often for Orbit chains there is only one?

src/decorators/getValidators.integration.test.ts
```diff
@@ -0,0 +1,75 @@
...
+const client = createPublicClient({
+  chain: nitroTestnodeL2,
+  transport: http(),
+}).extend(
+  rollupAdminLogicPublicActions({
+    rollup: l3Rollup,
+  }),
+);
+
+async function toggleValidator(validator: Address, state: boolean) {
```
fionnachan, line 22: this function takes a boolean for state, so it isn't toggling an internal state

a better name should be `setValidator`?

src/decorators/getValidators.integration.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    expect(initialValidators).toHaveLength(10);
+
+    await toggleValidator(randomAccount, false);
+    await toggleValidator(randomAccount, false);
+
+    const newValidators = await getValidators(client, { rollupAddress: l3Rollup });
+    // Setting the same validator multiple time to false doesn't add new validators
+    expect(newValidators).toEqual(initialValidators);
+
+    await toggleValidator(randomAccount, true);
```
fionnachan, line 51: this test shouldn't involve adding a validator after removing the validator

on another note, I'd also prefer breaking down this removal test to a few tests

1. remove an existing validator once
2. remove an existing validator twice
3. remove an address that is not a validator once
4. remove an address that is not a validator twice
5. add a validator and remove it

src/decorators/getValidators.integration.test.ts
chrstph-dvx: Default state when you start the nitro testnode

src/decorators/getValidators.integration.test.ts
chrstph-dvx: The test is mostly here to test case when you remove n times, add once at the end, to be sure it's considered added. Same for added multiple time, then removed once.
I don't think we would benefit from having both 1 and 2, or 3 and 4.

src/decorators/getValidators.integration.test.ts
fionnachan: can you update the comment please?

src/decorators/getValidators.integration.test.ts
fionnachan: i think performing an action (add/remove) with an existing validator vs a non-validator address is different so it'd be good to have both, but agreed that we can skip once vs twice cases

src/getValidators.unit.test.ts
```diff
@@ -11,9 +11,137 @@ const client = createPublicClient({
...
+  expect(validators).toEqual(['0x6a23CcC1c36D2aaA98AeF2a4471cf807DD22e45b']);
+  expect(isComplete).toBeTruthy();
+});
+
+it('getValidators return all validators with complete flag set to false', async () => {
+  const mockTransport = () =>
+    createTransport({
+      key: 'mock',
+      name: 'Mock Transport',
+      request: vi.fn(({ method, params }) => {
```
fionnachan, line 124: nit: why not name `params` as `param` so you don't have to do `param: params`? :D

src/getValidators.unit.test.ts
```diff
@@ -11,9 +11,137 @@ const client = createPublicClient({
...
+        return mockData({
+          logs: {
+            '0x448fdaac1651fba39640e2103d83f78ff4695f95727f318f0f9e62c3e2d77a10': validInput,
+            '0x6e29af776e4b08f92b484a3d4ecc506a4b6455bbd335a2547c4e97d6151f588c': '0xdeadbeef',
+          },
+          method,
+          param: params,
+        });
+      }) as any,
```
fionnachan, line 133: viem uses `as unknown as EIP1193RequestFn` so maybe we should do the same

https://github.com/wevm/viem/blob/main/src/clients/transports/createTransport.test.ts#L10C37-L10C68

src/getValidators.ts
```diff
@@ -0,0 +1,151 @@
...
+          data: tx.input,
+        });
+        return setValidators(acc, upgradeExecutorCall.args[1]);
+      }
+      case safeL2FunctionSelector: {
+        const safeCall = decodeFunctionData({
+          abi: [execTransactionABI],
+          data: tx.input,
+        });
+        return setValidators(acc, safeCall.args[2]);
```
fionnachan, line 137: it should be done through upgrade executor, so it's missing one step here

fionnachan (changes requested):

src/getValidators.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+export async function getValidators(
+  client: PublicClient,
+  { rollupAddress }: GetValidatorsParams,
+): Promise<GetValidatorsReturnType> {
+  const events = await client.getLogs({
+    address: rollupAddress,
+    event: ownerFunctionCalledEventAbi,
+    args: { id: 6n },
+    fromBlock: 0n,
```
spsjvc, line 98: if the rollup was deployed through the RollupCreator, you should be able to fetch the deployment tx hash with [createRollupFetchTransactionHash](https://github.com/OffchainLabs/arbitrum-orbit-sdk/blob/main/src/createRollupFetchTransactionHash.ts) and then fetch the transaction receipt and the block number

we should try/catch that and in case it fails fallback to `earliest`

```suggestion
    fromBlock: 0n,
```

src/getValidators.ts
```diff
@@ -0,0 +1,155 @@
...
+ *
+ * if (isComplete) {
+ *   // Validators were all fetched properly
+ * } else {
+ *   // Some validators might be missing
+ * }
+ */
+export async function getValidators(
+  client: PublicClient,
+  { rollupAddress }: GetValidatorsParams,
```
spsjvc, line 92: I think this is more in line with our other APIs?

```suggestion
  { rollup }: GetValidatorsParams,
```

src/getValidators.ts
```diff
@@ -0,0 +1,157 @@
...
+    abi,
+    data,
+  });
+  if (typeof args === 'undefined') {
+    return [];
+  }
+  return args;
+}
+
+function setValidators(acc: Set<Address>, input: Hex) {
```
spsjvc, line 44: this function name confused me when I first read it, can we update to

```suggestion
function updateAccumulator(acc: Set<Address>, input: Hex) {
```

spsjvc (approved):

src/getValidators.ts
```diff
@@ -0,0 +1,179 @@
...
+function getValidatorsFromFunctionData<
+  TAbi extends (typeof createRollupABI)[] | (typeof setValidatorABI)[],
+>({ abi, data }: { abi: TAbi; data: Hex }) {
+  const { args } = decodeFunctionData({
+    abi,
+    data,
+  });
+  if (typeof args === 'undefined') {
+    return [];
+  }
```
fionnachan, line 41: looks like `args` will never be `undefined` after you tightened up the type so this can be deleted

src/getValidators.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    switch (txSelectedFunction) {
+      case createRollupFunctionSelector: {
+        const [{ validators }] = getValidatorsFromFunctionData({
+          abi: [createRollupABI],
+          data: tx.input,
+        });
+
+        acc = new Set([...acc, ...validators]);
+        return acc;
```
fionnachan, line 144: I'd much prefer it without direct reassignment
```suggestion
        return new Set([...acc, ...validators]);
```

src/getValidators.unit.test.ts
```diff
@@ -0,0 +1,531 @@
...
+  http,
+} from 'viem';
+import { arbitrum } from 'viem/chains';
+import { it, expect, vi, describe } from 'vitest';
+import { getValidators } from './getValidators';
+import { rollupAdminLogicABI, safeL2ABI } from './abi';
+import { rollupAdminLogicPrepareFunctionData } from './rollupAdminLogicPrepareTransactionRequest';
+
+// Xai
+const rollupAdminLogicAddress = '0xc47dacfbaa80bd9d8112f4e8069482c2a3221336';
```
fionnachan, line 17: the rollup & rollup admin logic is on the same address

it is confusing that there are two different addresses for `rollupAdminLogicAddress` and `rollupAddress` in this file

I assume that is because they are the rollup addresses of 2 different Orbit chains

can you rename them and give some comments to them?

src/getValidators.unit.test.ts
```diff
@@ -0,0 +1,531 @@
...
+    const validator = '0xC0b97e2998edB3Bf5c6369e7f7eFfb49c36fA962';
+    const mockTransport = () =>
+      createTransport({
+        key: 'mock',
+        name: 'Mock Transport',
+        request: vi.fn(({ method, params }) => {
+          return mockData({
+            logs: {
+              '0x448fdaac1651fba39640e2103d83f78ff4695f95727f318f0f9e62c3e2d77a10':
+                safeSetValidatorHelper([[validator], [false]]),
```
fionnachan, line 405: we should run some tests with changing multiple validators in one tx

src/getValidators.ts
chrstph-dvx: Good catch

fionnachan (approved):

PR #114, opened 2024-06-05
feat: add new SequencerInbox getters

chrstph-dvx:
Part of FS-535

Getters:
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/144
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/145
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/146
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/147

Setters:
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/148
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/149
https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/150


src/actions/getMaxTimeVariation.ts
```diff
@@ -0,0 +1,33 @@
...
+export type GetMaxTimeVariationReturnType = {
+  delayBlocks: bigint;
+  futureBlocks: bigint;
+  delaySeconds: bigint;
+  futureSeconds: bigint;
+};
+
+export async function getMaxTimeVariation<TChain extends Chain | undefined>(
+  client: PublicClient<Transport, TChain>,
+  args: GetMaxTimeVariationParameters<false>,
```
fionnachan, line 20: does this need `<false>`?

src/actions/isBatchPoster.ts
```diff
@@ -0,0 +1,27 @@
...
+
+type Args = {
+  batchPoster: Address;
+};
+type SequencerInboxABI = typeof sequencerInbox.abi;
+export type IsBatchPosterParameters<Curried extends boolean = false> = ActionParameters<
+  Args,
+  'sequencerInbox',
+  Curried
+>;
```
fionnachan, line 13: this is very similar to `GetMaxTimeVariationParameters` and `IsValidKeysetHashParameters` etc

do we really want an independent one in every file?

chrstph-dvx: Yes, you can see the reasoning here: https://linear.app/offchain-labs/issue/FS-504/explore-ways-to-simplify-contract-specific-utilities-for-v1#comment-6ca0cf13

src/actions/getMaxTimeVariation.ts
chrstph-dvx: good point, we don't

src/types/Actions.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export type ActionParameters<Args, ContractName extends string, Curried extends boolean> = Prettify<
+  Curried extends false
+    ? Args & { [key in ContractName]: Address }
+    : Args extends Record<string, never>
+    ?
+        | {
+            [key in ContractName]: Address;
+          }
+        | void
```
fionnachan, line 18: continuing our previous discussion, not very sure about using `void` for optional objects but approving this as it works

fionnachan (approved):

src/types/Actions.ts
chrstph-dvx: It's slightly different here. If we use `void, we allow `getMaxTimeVariation()` if the sequencerInbox was passed to the decorator previously. If we use `{ [key in ContractName]?: Address }` instead, then we always require a parameter, also `getMaxTimeVariation({ sequencerInbox: undefined })` would now be a valid option.

PR #124, opened 2024-06-17
feat: add getBatchPosters

chrstph-dvx:
Closes FS-595


src/getBatchPosters.ts
```diff
@@ -0,0 +1,181 @@
...
+  const events = createRollupTransactionHash
+    ? [{ transactionHash: createRollupTransactionHash }, ...sequencerInboxEvents]
+    : sequencerInboxEvents;
+  const txs = await Promise.all(
+    events.map((event) =>
+      publicClient.getTransaction({
+        hash: event.transactionHash,
+      }),
+    ),
+  );
```
chrstph-dvx, line 133: This is the main difference, we need to get `createRollup` event data from `rollup` contract, but rest of the logs from `sequencerInbox`

fionnachan (approved):
LGTM

PR #129, opened 2024-06-20
feat: add getKeysets

chrstph-dvx:
Closes fs-586


src/getValidKeysets.ts
```diff
@@ -0,0 +1,109 @@
...
+    events: [SetValidKeysetEventAbi, InvalidateKeysetEventAbi],
+    fromBlock: blockNumber,
+    toBlock: 'latest',
+  });
+
+  let isAccurate = true;
+  const keysets = events.reduce((acc, event) => {
+    switch (event.eventName) {
+      case SetValidKeysetEventAbi.name: {
+        const { keysetHash, keysetBytes } = event.args;
```
fionnachan, line 83: we need cases for Upgrade Executor & Safe too because users are very likely calling through Upgrade Executor, and probably using Safe

fionnachan (dismissed):

src/getValidKeysets.ts
fionnachan: After discussion, @chrstph-dvx clarified that the event args are the data we wanted so we do not need to look into the tx calldata

src/getKeysets.ts
```diff
@@ -0,0 +1,109 @@
...
+import { createRollupFetchTransactionHash } from './createRollupFetchTransactionHash';
+
+const SetValidKeysetEventAbi = getAbiItem({ abi: sequencerInboxABI, name: 'SetValidKeyset' });
+const InvalidateKeysetEventAbi = getAbiItem({ abi: sequencerInboxABI, name: 'InvalidateKeyset' });
+
+export type GetKeysetsParams = {
+  /** Address of the rollup we're getting list of keysets from */
+  rollup: Address;
+  /** Address of the sequencerInbox we're getting logs from */
+  sequencerInbox: Address;
```
TucksonDev, line 12: We don't need to pass the sequencerInbox as parameter, since we can obtain it from the Rollup address.

You can either call the `sequencerInbox` method in the rollupContract, or use the `createRollupFetchCoreContracts` method of the SDK.

src/getKeysets.ts
```diff
@@ -0,0 +1,109 @@
...
+      rollup,
+      publicClient,
+    });
+    const receipt = await publicClient.waitForTransactionReceipt({
+      hash: createRollupTransactionHash,
+    });
+    blockNumber = receipt.blockNumber;
+  } catch (e) {
+    console.warn(`[getKeysets] ${(e as any).message}`);
+    blockNumber = 'earliest';
```
TucksonDev, line 69: We haven't generalized this, but we also have the `earliestRollupCreatorDeploymentBlockNumber` constant in `createRollupFetchTransactionHash` that we could use for these cases.

src/getKeysets.ts
```diff
@@ -0,0 +1,109 @@
...
+   * If logs contain unknown signature, keysets list might:
+   * - contain false positives (keysets that were removed, but returned as keyset)
+   * - contain false negatives (keysets that were added, but not present in the list)
+   */
+  isAccurate: boolean;
+  /** Map of keyset hash to keyset bytes
+   *  keyset hash are used to invalidate a given keyset
+   */
+  keysets: {
+    [keysetHash: Hex]: Address;
```
TucksonDev, line 25: nitpick: keyset bytes are not really an Address, but a number of bytes.

TucksonDev (changes requested):
Thanks @chrstph-dvx ! This looks very good 🚀 

The only requested change is removing the sequencerInbox address as a required parameter.

TucksonDev (approved):
LGTM!
Thanks @chrstph-dvx !! 🚀

PR #131, opened 2024-06-25
refactor: Add fallback to core contracts fetch

chrstph-dvx:
If createRollupFetchCoreContracts initially fails to retrieve contracts from the receipt, default to multicall to fetch the contracts

closes FS-614

Setting the PR as draft for now, we are unable to fetch:
* `adminProxy`
* `deployedAtBlockNumber`
* `upgradeExecutor`

And changing type to mark those 3 properties optional would be a breaking change, so I'm setting the PR to draft for now


spsjvc:
Not planned.

PR #147, opened 2024-07-10
feat: Add v1 upgradeExecutor getters

chrstph-dvx:
part of FS-658

### DIFF
--- a/src/actions/hasRole.ts
+++ b/src/actions/hasRole.ts
@@ -1,0 +1,27 @@
+import { Address, Chain, PublicClient, ReadContractReturnType, Transport } from 'viem';
+import { upgradeExecutorABI } from '../contracts/UpgradeExecutor';
+import { UpgradeExecutorRole } from '../upgradeExecutorEncodeFunctionData';
+import { ActionParameters } from '../types/Actions';
+
+export type HasRoleParameters<Curried extends boolean = false> = ActionParameters<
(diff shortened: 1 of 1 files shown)


src/actions/hasRole.ts
```diff
@@ -0,0 +1,29 @@
...
+  Args,
+  'upgradeExecutor',
+  Curried
+>;
+
+export type HasRoleReturnType = ReadContractReturnType<UpgradeExecutorABI, 'hasRole'>;
+
+export async function hasRole<TChain extends Chain | undefined>(
+  client: PublicClient<Transport, TChain>,
+  args: HasRoleParameters,
```
chrstph-dvx, line 21: This will be changed once we merge https://github.com/OffchainLabs/arbitrum-orbit-sdk/pull/140

Then it will use the same mechanism as rollupAdmin getters where we either accept `upgradeExecutor` or `sequencerInbox` and fetch `upgradeExecutor` contract if needed

src/actions/hasRole.ts
fionnachan, line 1: other than `hasRole`, i think we also want to include `upgradeExecutorFetchPrivilegedAccounts`?

fionnachan: and all these should be here?

image

fionnachan (approved):

PR #149, opened 2024-07-10
feat: Add v1 ArbOwnerPublic setters

chrstph-dvx:
part of FS-529


src/actions/setParentPricingRewardRecipient.ts
```diff
@@ -0,0 +1,43 @@
...
+  const validatedPublicClient = validateChildChainPublicClient(client);
+  const { account, upgradeExecutor, recipient } = params;
+
+  const request = await client.prepareTransactionRequest({
+    chain: client.chain,
+    account,
+    ...withUpgradeExecutor({
+      to: arbOwner.address,
+      upgradeExecutor,
+      args: [recipient],
```
fionnachan, line 36: i am wondering if we would like to check the `Address` type in this function, or outside? currently it is set up to be done outside of this function

what are your thoughts?

PR #159, opened 2024-08-08
feat: handle generating abi from proxy

spsjvc:


src/contracts.ts
```diff
@@ -47,163 +48,6 @@ export const upgradeExecutor = {
   ]),
 };
 
-const tokenBridgeCreatorABI = [
```
spsjvc, line 50: replaced the hardcoded ABI with the auto-generated ABI

patches/@wagmi+cli+1.5.2.patch
spsjvc, line 1: these changes are necessary so we can keep the generated addresses be proxies, but the generated abi be used from the implementation addresses

chrstph-dvx (approved):

PR #167, opened 2024-08-14
feat: update wasm utils as consensus release utils

spsjvc:
The original approach made sense when the only info on the consensus release was `wasmModuleRoot`. In a future PR, I'll be adding `maxSupportedArbOSVersion` to each consensus release for extra validation, so having utils to return the whole object instead of just a single property makes more sense for future-proofing.

### DIFF
--- a/src/wasmModuleRoot.ts
+++ b/src/wasmModuleRoot.ts
@@ -1,6 +1,6 @@
 import { Hex } from 'viem';
 
-export type ConsensusRelease = {
+type ConsensusRelease = {
   version: number;
   wasmModuleRoot: Hex;
 };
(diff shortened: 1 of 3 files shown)


src/wasmModuleRoot.ts
```diff
@@ -58,40 +58,40 @@ type ConsensusReleases = typeof consensusReleases;
...
-export function getConsensusVersion<TWasmModuleRoot extends WasmModuleRoot>(
+export function getConsensusReleaseByWasmModuleRoot<TWasmModuleRoot extends WasmModuleRoot>(
   wasmModuleRoot: TWasmModuleRoot,
-): GetConsensusVersion<TWasmModuleRoot> {
-  const consensusVersion = consensusReleases
+): GetConsensusReleaseByWasmModuleRoot<TWasmModuleRoot> {
+  const consensusRelease = consensusReleases
     //
-    .find((release) => release.wasmModuleRoot === wasmModuleRoot)!.version;
+    .find((release) => release.wasmModuleRoot === wasmModuleRoot)!;
```
fionnachan, line 86: you can remove the non-null assertion now

fionnachan (approved):

PR #177, opened 2024-09-03
Configure Fast Withdrawals using Safe

InoMurko:
Rollups tend to have ownership transferred to a Safe multi-signature account.  These 4 scripts propose transactions to the Safe if there's at least one EOA account in the Safe ownership.


src/createSafePrepareTransactionRequest.ts
```diff
@@ -77,7 +77,7 @@ export const SafeProxyFactoryAbi = [
  */
 export type CreateSafePrepareTransactionRequestParams<TChain extends Chain | undefined> = {
   publicClient: PublicClient<Transport, TChain>;
-  account: PrivateKeyAccount;
+  account: PrivateKeyAccount | `0x${string}`;
```
spsjvc, line 80: let's just switch to an account address and drop `PrivateKeyAccount` (and update jsdoc accordingly)

```suggestion
  account: Address;
```

src/setAnyTrustFastConfirmerPrepareTransactionRequest.ts
```diff
@@ -17,7 +17,7 @@ export type SetAnyTrustFastConfirmerPrepareTransactionRequestParams<
   TChain extends Chain | undefined,
 > = {
   publicClient: PublicClient<Transport, TChain>;
-  account: PrivateKeyAccount;
+  account: PrivateKeyAccount | `0x${string}`;
```
spsjvc, line 20: let's just switch to an account address and drop `PrivateKeyAccount` (and update jsdoc accordingly)

```suggestion
  account: Address;
```

src/createSafePrepareTransactionRequest.ts
InoMurko: good idea!

PR #178, opened 2024-09-03
test: add integration tests for new SequencerInbox utils

chrstph-dvx:


src/actions/sequencerInbox.integration.test.ts
```diff
@@ -0,0 +1,220 @@
...
+    const isRandomAddressBatchPosterAfterDisabling = await isBatchPoster(client, {
+      sequencerInbox: l3SequencerInbox,
+      batchPoster: randomAddress,
+    });
+    expect(isRandomAddressBatchPosterAfterDisabling).toBeFalsy();
+  });
+});
+
+describe('keyset management', () => {
+  async function deployAnytrustRollupWithKeyset(keyset: Hex) {
```
spsjvc, line 139: a chain can't be rollup _and_ anytrust
```suggestion
  async function deployAnyTrustChainWithKeyset(keyset: Hex) {
```

src/actions/__snapshots__/sequencerInbox.integration.test.ts.snap
```diff
@@ -0,0 +1,10 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`max time variation management > getMaxTimeVariation successfully fetches max time variation 1`] = `
```
spsjvc, line 3: i don't think it's worth adding a snapshot for a small test case like this, let's just inline

spsjvc (approved):

PR #179, opened 2024-09-09
refactor: split getters into separate files

spsjvc:
In a follow-up PR I'll be adding unit tests for each individual getter, so made sense to split them into multiple files.


src/utils/getBlockExplorerUrl.ts
fionnachan, line 1: do you think this function should be exported from `utils/index.ts`?

src/createRollupPrepareTransactionRequest.ts
```diff
@@ -8,7 +8,7 @@ import { validateParentChain } from './types/ParentChain';
 import { isCustomFeeTokenAddress } from './utils/isCustomFeeTokenAddress';
 import { ChainConfig } from './types/ChainConfig';
 import { isAnyTrustChainConfig } from './utils/isAnyTrustChainConfig';
-import { getRollupCreatorAddress } from './utils/getters';
+import { getRollupCreatorAddress } from './utils/getRollupCreatorAddress';
```
fionnachan, line 11: could have imported from `./utils`, is being more specific the preference here?

PR #181, opened 2024-09-09
feat: custom parent chains

spsjvc:
Closes FS-800.


src/testHelpers.ts
```diff
@@ -179,3 +179,34 @@ export async function createRollupHelper({
     createRollupInformation,
   };
 }
+
+export function testHelper_createCustomParentChain(params?: { id?: number }) {
```
spsjvc, line 183: opted in to use a `testHelper_` prefix because it's easier to differentiate what's a test helper and what's actual production code when writing a test case

chrstph-dvx (approved):

PR #185, opened 2024-09-11
refactor: use getValidators method in set validators example

Jason-W123:


examples/set-new-validators/index.ts
fionnachan, line 1: image

Even though I am using my API key with infura, it still throws 429 (too many requests)

In `getValidators`, we are getting the event logs, then get the transaction, so there are quite some calls.

Perhaps you can add a wait before calling the method the second time?

fionnachan (approved):

PR #252, opened 2025-03-27
fix: update rpc and api key for abi generation

spsjvc:


.env.example
```diff
@@ -1,5 +1,6 @@
 ETHERSCAN_API_KEY=
 ARBISCAN_API_KEY=
+ARBISCAN_NOVA_API_KEY=
```
spsjvc, line 3: Arbiscan now has a separate key for the Nova explorer

wagmi.config.ts
```diff
@@ -36,7 +36,17 @@ function sleep(ms: number = 1_000) {
...
 }
 
+// @ts-ignore
+mainnet.rpcUrls.default.http[0] = 'https://mainnet.gateway.tenderly.co';
+// @ts-ignore
+mainnet.rpcUrls.public.http[0] = 'https://mainnet.gateway.tenderly.co';
+// @ts-ignore
+sepolia.rpcUrls.default.http[0] = 'https://sepolia.gateway.tenderly.co';
+// @ts-ignore
+sepolia.rpcUrls.public.http[0] = 'https://sepolia.gateway.tenderly.co';
```
spsjvc, line 46: This is because both `https://cloudflare-eth.com` and `https://rpc.sepolia.org` have issues.
I know it's a bit dirty, but we very rarely run this (only when Nitro contracts or ArbOS is bumped), and this is a quick fix.

fionnachan (approved):
tested and it worked for me

PR #253, opened 2025-03-27
feat: add getters for defaults

spsjvc:


src/getDefaultValidatorAfkBlocks.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (parentChainIsCustom) {
+    throw new Error(
+      `[getDefaultValidatorAfkBlocks] can't provide defaults for custom parent chain with id ${parentChainId}`,
+    );
+  }
+
+  const blocksPerMinute = 60 / getParentChainBlockTime(parentChainId);
+
+  // 28 days
```
TucksonDev, line 20: Should we add a comment here explaining why the default is 28 days?
https://github.com/OffchainLabs/nitro-contracts/blob/main/src/rollup/RollupAdminLogic.sol#L58-L60

Or in some other place, not sure. The other amounts are more or less understandable, this one I feel is the only one that seems to deserve further explanation.

TucksonDev (approved):
LGTM. Left a small non-blocking nit

PR #266, opened 2025-04-29
refactor: use etherscan api v2

spsjvc:
Closes FS-1240.

https://docs.etherscan.io/etherscan-v2


fionnachan (approved):
tested and LGTM, can you run `yarn generate` on this PR or after this is merged? I see some very minor changes in the generated files

another thing to note: `arbOwnerPublicABI` changes quite a bit on Arb Sepolia, so if we want to keep old version support we need versioning

PR #272, opened 2025-05-09
feat: update nitro contracts to v2.1.3

spsjvc:
Closes FS-1249.


src/contracts/RollupCreator/v2.1.ts
spsjvc, line 1: ABIs stayed the same, changes were internal

fionnachan (approved):
cross-checked, LGTM

PR #274, opened 2025-05-15
chore: adds dependabot.yml

douglance:


.github/dependabot.yml
```diff
@@ -0,0 +1,71 @@
...
+    ignore:
+      - dependency-name: '@arbitrum/*'
+      - dependency-name: '@offchainlabs/*'
+    # Disable version updates for external dependencies - security only
+    open-pull-requests-limit: 0
+    reviewers:
+      - 'OffchainLabs/full-stack'
+    labels:
+      - 'dependencies'
+      - 'external'
```
fionnachan, line 71: since we have disabled opening PRs for external deps, i don't think we need these?

.github/dependabot.yml
```diff
@@ -0,0 +1,71 @@
...
```
fionnachan, line 36: similarly

.github/dependabot.yml
```diff
@@ -0,0 +1,71 @@
+version: 2
+updates:
+  # === Root package.json (/) ===
+  # Internal dependencies
+  - package-ecosystem: 'npm'
+    directory: '/'
```
fionnachan, line 6: can't you use `directories` instead and remove the duplicate?

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--

.github/dependabot.yml
```diff
@@ -0,0 +1,71 @@
...
+        update-types: ['version-update:semver-patch']
+    reviewers:
+      - 'OffchainLabs/full-stack'
+    labels:
+      - 'dependencies'
+      - 'internal'
+
+  # External dependencies
+  - package-ecosystem: 'npm'
+    directory: '/'
```
fionnachan, line 24: `directories`?

PR #487, opened 2025-07-28
feat: add versioning to deployment config defaults

spsjvc:


src/createRollupPrepareDeploymentParamsConfigDefaults.ts
```diff
@@ -29,3 +38,22 @@ export const defaults = {
...
+): typeof defaultsV2Dot1;
+export function createRollupPrepareDeploymentParamsConfigDefaults(
+  rollupCreatorVersion: 'v3.1',
+): typeof defaultsV3Dot1;
+export function createRollupPrepareDeploymentParamsConfigDefaults(
+  rollupCreatorVersion?: never,
+): typeof defaultsV3Dot1;
+export function createRollupPrepareDeploymentParamsConfigDefaults(
+  rollupCreatorVersion: 'v3.1' | 'v2.1' = 'v3.1',
+) {
```
spsjvc, line 53: made it into a function so we can version it + added overloads so return type is based on provided version

PR #492, opened 2025-08-06
feat: add versioning to deployment config

spsjvc:


src/createRollupPrepareDeploymentParamsConfig-v3.1.ts
spsjvc, line 1: This is just moved from `src/createRollupPrepareDeploymentParamsConfig.ts`

PR #500, opened 2025-08-09
feat: add versioning to deployment params config

spsjvc:


vitest.type.config.ts
```diff
@@ -0,0 +1,16 @@
+import { configDefaults, defineConfig, mergeConfig } from 'vitest/config';
```
chrstph-dvx, line 1: Why do we need this file? We have test type running in unit test already.
`arbOwnerPrepareTransactionRequest`

```
expectTypeOf<typeof client.arbOwnerPrepareTransactionRequest<'addChainOwner'>>().toBeCallableWith(
    {
      functionName: 'addChainOwner',
      args: [randomAccount.address],
      upgradeExecutor: false,
      account: randomAccount.address,
    },
  );
```

chrstph-dvx (approved):

PR #614, opened 2025-09-24
feat: deploy c2p router on op chain

yahgwai:


TucksonDev (approved):
LGTM. Left a few minor non-blocking comments.

PR #638, opened 2026-01-23
feat: default to ArbOS 51

spsjvc:


Jason-W123 (approved):
Local tests passed, can start the chain and confirm the assertion without errors.

PR #662, opened 2026-03-19
feat: deploy tokenbridge without orbitchainclient

yahgwai:
Currently in order to deploy contracts for a chain the flow is:
1. deploy contracts
2. start the node
3. deploy the tokenbridge

However it's actually possible to do
1. deploy contracts
2. deploy tokenbridge
3. start the node

When the node starts up the first thing it will do is process retryables that were created in step 2 - we dont actually need the node started in order to create those retryables which will be redeemed when the node starts. The current create*tokenbridge methods in the sdk rely on the node having started in order to estimate gas price of the retryable, but we can just assume the gas price is 0.1gwei and hardcode something higher to guarantee the retryables will succeed. That's what we do here, and remove any dependency on having the node started in order to deploy the tokenbridge.


src/calculateRetryableSubmissionFee.ts
```diff
@@ -0,0 +1,28 @@
+import { Address, PublicClient, Transport, Chain, parseAbi } from 'viem';
+
+// Matches DEFAULT_SUBMISSION_FEE_PERCENT_INCREASE in @arbitrum/sdk's ParentToChildMessageGasEstimator
+const SUBMISSION_FEE_PERCENT_INCREASE = 300n;
```
TucksonDev: Should this be in the `constants` file?

src/createTokenBridgePrepareTransactionRequest.ts
```diff
@@ -39,46 +45,110 @@ export type CreateTokenBridgePrepareTransactionRequestParams<
...
-}: CreateTokenBridgePrepareTransactionRequestParams<TParentChain, TOrbitChain>) {
+}: CreateTokenBridgePrepareTransactionRequestParams<TParentChain>) {
   const { chainId } = validateParentChain(parentChainPublicClient);
 
-  const isTokenBridgeAlreadyDeployed = await isTokenBridgeDeployed({
-    parentChainPublicClient,
-    orbitChainPublicClient,
-    rollup: params.rollup,
-    tokenBridgeCreatorAddressOverride,
+  const tokenBridgeCreatorAddress =
```
TucksonDev, line 59: We could keep this logic inside `isTokenBridgeAlreadyDeployed` but make `orbitChainPublicClient` optional, and only check the L2 if it's present. That way we don't duplicate the code here.
What do you think?

src/createTokenBridgePrepareTransactionRequest.ts
```diff
@@ -39,46 +45,110 @@ export type CreateTokenBridgePrepareTransactionRequestParams<
...
+  // Parent-chain-only idempotency check.
+  // If router is non-zero, a prior createTokenBridge tx has executed on the parent chain.
+  // If the L2 retryables failed, they should be manually redeemed rather than redeployed.
+  const [router] = await parentChainPublicClient.readContract({
+    address: tokenBridgeCreatorAddress,
+    abi: tokenBridgeCreatorABI,
+    functionName: 'inboxToL2Deployment',
+    args: [inbox],
+  });
+  if (router !== zeroAddress) {
```
TucksonDev, line 77: Additionally, the error message below this check is not 100% accurate now, because the retryable ticket might have failed (so no code in the router address), but the L1 will have stored the expected address in `router`.

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
```diff
@@ -1,26 +1,27 @@
...
+import { calculateRetryableSubmissionFee } from './calculateRetryableSubmissionFee';
 
 export type TransactionRequestRetryableGasOverrides = {
   gasLimit?: GasOverrideOptions;
   maxFeePerGas?: GasOverrideOptions;
   maxSubmissionCost?: GasOverrideOptions;
 };
 
-export type CreateTokenBridgePrepareRegisterWethGatewayTransactionRequestParams<
+export type CreateTokenBridgePrepareSetWethGatewayTransactionRequestParams<
```
TucksonDev, line 24: This name is clearer, but calling out that we would be changing this exported type. Since we're adding breaking changes, this PR is probably a good place to do that though.

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
```diff
@@ -133,21 +123,18 @@ export async function createTokenBridgePrepareSetWethGatewayTransactionRequest<
...
   });
 
-  // get token bridge contracts
   const tokenBridgeContracts = await createTokenBridgeFetchTokenBridgeContracts({
     inbox,
     parentChainPublicClient,
     tokenBridgeCreatorAddressOverride,
   });
 
-  // check whether the weth gateway is already registered in the router
```
TucksonDev, line 150: I'd keep the comments in this file. The code is mostly clear enough, but it doesn't hurt to have the additional comment, specially in some of them (like this one)

src/createTokenBridgePrepareTransactionRequest.ts
```diff
@@ -39,46 +45,110 @@ export type CreateTokenBridgePrepareTransactionRequestParams<
...
+          percentIncrease: retryableGasOverrides.maxSubmissionCostForContracts.percentIncrease,
+        })
+      : submissionCostForContractsEstimate;
+
+  const maxGasForContracts =
+    retryableGasOverrides && retryableGasOverrides.maxGasForContracts
+      ? applyPercentIncrease({
+          base:
+            retryableGasOverrides.maxGasForContracts.base ??
+            createTokenBridgeDefaultMaxGasForContracts,
```
TucksonDev, line 143: iirc, we are already adding a 30% increase in the default, so if the overrides don't include a `base`, but only a `percentIncrease` it will increase that percentage on top of the already 30% increase. I'd say we should either apply it if no overrides are present (after the `:` in this operation), or comment that here.

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
```diff
@@ -177,45 +163,41 @@ export async function createTokenBridgePrepareSetWethGatewayTransactionRequest<
...
+    calldataSize,
   );
 
   //// apply gas overrides
   const gasLimit =
     retryableGasOverrides && retryableGasOverrides.gasLimit
       ? applyPercentIncrease({
-          base: retryableGasOverrides.gasLimit.base ?? retryableTicketGasEstimates.gasLimit,
+          base:
+            retryableGasOverrides.gasLimit.base ?? createTokenBridgeDefaultGasLimitForWethGateway,
```
TucksonDev, line 179: [This comment ](https://github.com/OffchainLabs/arbitrum-chain-sdk/pull/662/changes#r3063745534) also applies here.

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
```diff
@@ -228,17 +210,13 @@ export async function createTokenBridgePrepareSetWethGatewayTransactionRequest<
...
   });
 
-  // prepare the transaction request with a call to the upgrade executor
   // @ts-expect-error -- todo: fix viem type issue
   const request = await parentChainPublicClient.prepareTransactionRequest({
     chain: parentChainPublicClient.chain,
     to: rollupCoreContracts.upgradeExecutor,
     data: upgradeExecutorEncodeFunctionData({
       functionName: 'executeCall',
-      args: [
```
TucksonDev, line 238: I'd keep the comments that identify the parameters of the `executeCall` function

src/createTokenBridge-ethers.ts
TucksonDev, line 1: Now that we don't need `ethers` in this file (the current calls can be made in viem), we could rename this file to something different.
For context, this file existed here because we needed ethers' Providers to call the ArbSDK for estimations, which is not the case anymore.

One idea is to rename it to something related to the gas parameters for createTokenBridge, or moving `getFactoryDeploymentDataSize` to the `createTokenBridgePrepareTransactionRequest` file, and rename this to `createTokenBridge-getContractsDeploymentData` or something like that.

src/createTokenBridgePrepareTransactionRequest.ts
yahgwai: I think it's better not to, in another comment you mention the innacurate error message from a failed deployment, if we added the optional param to `isTokenBridgeAlreadyDeployed` that would make it named incorrectly in a similar way

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
yahgwai: hmmm. good point and good call out. @spsjvc you happy with this rename? might be unnecessary even if clearer

src/createTokenBridge-ethers.ts
yahgwai: Hmm, maybe, but we still use ethers in this file quite a bit - even though we could migrate. maybe we leave it as is for now and move to viem + rename in a different pr?

src/calculateRetryableSubmissionFee.ts
yahgwai: updated

src/createTokenBridgePrepareTransactionRequest.ts
yahgwai: Updated

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
yahgwai: I left some of the comments which were non obvious - but I think removing this one is better given that it's obvious from the code. I prefer comments that add something to the code, or describe complicated code, rather than repear the code

I left this one:
> // (we first encode dummy data, to get the retryable message estimates)

src/createTokenBridgePrepareTransactionRequest.ts
yahgwai: This particular one comes from the hardcoded constant - so not being applied already

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
yahgwai: Also this, one comes from a constant.

However, we were double applying for max submission costs though, so good catch there! Fixed that

src/createTokenBridgePrepareTransactionRequest.ts
TucksonDev: That's true, let's keep it here.

src/createTokenBridgePrepareTransactionRequest.ts
TucksonDev: I think the hardcoded constant already has the 30% increase (judging by the comment `~30% headroom over observed gas usage for token bridge retryables`), but tbh, I don't remember what were the actual limits.

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
TucksonDev: Sounds good

src/createTokenBridge-ethers.ts
TucksonDev: Sounds good

TucksonDev (approved):
LGTM 🚀

src/createTokenBridgePrepareSetWethGatewayTransactionRequest.ts
spsjvc: that sounds good to me!

src/createTokenBridgePrepareTransactionRequest.ts
```diff
@@ -39,46 +46,117 @@ export type CreateTokenBridgePrepareTransactionRequestParams<
...
-  orbitChainPublicClient,
   account,
   gasOverrides,
   retryableGasOverrides,
   tokenBridgeCreatorAddressOverride,
-}: CreateTokenBridgePrepareTransactionRequestParams<TParentChain, TOrbitChain>) {
+}: CreateTokenBridgePrepareTransactionRequestParams<TParentChain>) {
   const { chainId } = validateParentChain(parentChainPublicClient);
 
-  const isTokenBridgeAlreadyDeployed = await isTokenBridgeDeployed({
```
spsjvc, line 54: The file `src/createTokenBridge.ts` also uses `isTokenBridgeDeployed` but it hasn't been changed there. Is it still necessary given it uses `createTokenBridgePrepareTransactionRequest` internally?

We might be able to get rid of `isTokenBridgeDeployed` completely.

yahgwai: agreed, deleted

PR #687, opened 2026-04-08
feat: get chain contract versions

alxdca:

* **#750**
* **#748**
* **#711**
* **#746**
* **#709**
* **#701**
* ➡️ **#687**
* `main`
</details></td></tr></table>

## Summary

- Add `getChainContractVersions` to discover deployed Nitro contract versions for an Orbit chain.
- Use the native `@arbitrum/chain-actions` versioner and return version details with any upgrade recommendation.
- Export the helper from the SDK and add unit coverage for result handling.
- Add a runnable example using an inbox address and parent-chain RPC URL.

## Why / Context

Orbit chain operators need a straightforward way to inspect their deployed contract versions and determine whether an upgrade is recommended. This exposes the native chain versioner through the chain SDK and documents an end-to-end usage path.


src/getChainContractVersions.ts
```diff
@@ -0,0 +1,43 @@
...
+    !('versions' in parsed) ||
+    !('upgradeRecommendation' in parsed)
+  ) {
+    throw new Error('Failed to parse Orbit chain contract versions');
+  }
+
+  return parsed as GetChainContractVersionsResult;
+}
+
+export async function getChainContractVersions(
```
spsjvc: We could be a bit more precise with the name:

```suggestion
export async function getNitroContractsVersions(
```

spsjvc (approved):

PR #690, opened 2026-04-08
feat(scripting): add zod schemas and viem transform helpers

yahgwai:
## Summary
Adds zod schemas and transforms for every SDK function that takes RPC/client input. Schemas validate JSON input and transforms convert it to SDK-ready arguments -- RPC URLs become viem clients, private keys become accounts, strings become bigints. The target audience here is devs who are already familiar with zod, and helping them to write scripts without messing up the arg parsing, and letting them follow a opinionated structure. Devs who arent familiar with zod dont get benefit here, but it's not a complicated library to learn, and I think it's a useful thing to know about.

WARNING: we're creating schemas for every function, so there's a lot of code being added here, but two things are worth focusing on which help us to gain confidence in the correctness of the code:
* The schema type tests - here we're trying to test that the output of the transformed schema could be directly into its equivalent sdk function, so if you're confident those tests work then in theory the schemas are good from a correctness point of view
* In all cases where we can we set the return type of the schema transform to Parameters

### Not included for now
* Additional parsing libs. Now that we have zod schemas which will parse json, we can use existing tooling to convert yaml, env vars or cli args to json and than parse with the same schema giving scripts more options of what they expose to users.
* Describe() - we can add documentation to the schemas to give consumers and idea of what values to put into the params. Those descriptions can be accessed via a dry-run/help command or zod can be converted into json schemas to directly expose the documentation.
* Documentation and readme. Three sets of documentation is required: one for sdk maintainers, one for script writers and one for users of the cli.
* Extending the CLI with some standard useful scripts to be used by SRE/other consumers


src/scripting/schemas/common.ts
```diff
@@ -0,0 +1,140 @@
+import { z } from 'zod';
+
+export const hexSchema = z
+  .string()
+  .regex(/^0x[0-9a-fA-F]*$/, 'Invalid hex string')
+  .transform((val) => val as `0x${string}`);
+
+export const addressSchema = z
+  .string()
+  .regex(/^0x[0-9a-fA-F]{40}$/, 'Invalid Ethereum address')
```
douglance: viem has [isHex](https://v1.viem.sh/docs/utilities/isHex.html#ishex), [isAddress](https://v1.viem.sh/docs/utilities/isAddress.html), etc. predicates we can use

package.json
douglance, line 1: are we sure we need these?

src/scripting/schemas/common.ts
yahgwai: agreed, updated

package.json
yahgwai: no, removed

douglance (approved):

PR #705, opened 2026-04-24
feat: misc changes to the cli and deployFullChain

yahgwai:
Misc changes for the scripting CLI: a Docker-publishable binary wrapping the full command registry plus high-level
deployment workflows.

- Dockerfile + docker-publish workflow publishing offchainlabs/arbitrum-chain-sdk:cli-rc images to Docker Hub
- Workflow schema input rearranged grouped into createRollupParams / tokenBridgeParams / ownershipTransferParams, with inputSchema exports
for field reuse across workflows
- deployFullChain returns a ChainDeploymentInfo-shaped output
- scriptUtils: JSONC input, @<file> read-from-disk, - read-from-stdin, -o <file> write output, --schema to emit the JSON
Schema for a command
- CLI progress logs routed to stderr so stdout is reserved for the structured result
- Coverage overrides extended for workflow commands (nodeConfigParams, rollupDeploymentBlockNumber)
- Make all schemas deep strict
- Add inbox allow list setting to deploy full chain
- Add refine checks for custom genesis
- Added skip functionality to schema coverage tests
- Allow genesis.chainconfig to be passed in to the chain config directly


douglance:
1. `src/scripting/schemas/common.ts`, line 168

Could we avoid adding `arbitrum` keys here when the caller did not supply them? It looks like the tunable-only shape can produce own properties with `undefined` values, which then get spread over the defaults in `prepareChainConfig`.

Would building this object only from supplied fields preserve defaults like `InitialArbOSVersion`, `MaxCodeSize`, and `MaxInitCodeSize`?

2. `src/scripting/workflows/deployFullChain.ts`, line 341

Should this use the chain config from the generated deployment config when `nodeConfigParams` is supplied? It looks like callers can omit an explicit `chainConfig`, still get a generated chain config for deployment, but then skip `prepareNodeConfig` because this local `chainConfig` is undefined.

I think we should either derive the node config and returned `chainId` / `chainConfig` from `config.chainConfig`, or validate that node config generation requires an explicit chain config.

3. `src/scripting/schemas/createRollupPrepareDeploymentParamsConfig.ts`, line 91

Could this custom-genesis check compare against the whole default assertion state? It looks like changes to fields like `endHistoryRoot`, `machineStatus`, or `u64Vals[1]` would still be a non-default assertion state, but would skip the `dataCostEstimate` / `chainConfig` validation here.

Would a full default-state comparison be safer?

yahgwai:
> src/scripting/schemas/common.ts, line 168
Could we avoid adding arbitrum keys here when the caller did not supply them? It looks like the tunable-only shape can produce own properties with undefined values, which then get spread over the defaults in prepareChainConfig.
 Would building this object only from supplied fields preserve defaults like InitialArbOSVersion, MaxCodeSize, and MaxInitCodeSize?

Fixed: rewrote the transform to use destructure-rest, dropping validator-only literals and only forwarding tunable keys the caller actually supplied (no undefined own-props), so prepareChainConfig defaults are preserved.

> src/scripting/workflows/deployFullChain.ts, line 341
Should this use the chain config from the generated deployment config when nodeConfigParams is supplied? It looks like callers can omit an explicit chainConfig, still get a generated chain config for deployment, but then skip prepareNodeConfig because this local chainConfig is undefined.
I think we should either derive the node config and returned chainId / chainConfig from config.chainConfig, or validate that node config generation requires an explicit chain config.

Fixed: hoisted the chainConfig default into the transform so the local chainConfig is always defined and matches the deployed one; prepareNodeConfig now runs whenever nodeConfigParams is supplied.

> src/scripting/schemas/createRollupPrepareDeploymentParamsConfig.ts, line 91
Could this custom-genesis check compare against the whole default assertion state? It looks like changes to fields like endHistoryRoot, machineStatus, or u64Vals[1] would still be a non-default assertion state, but would skip the dataCostEstimate / chainConfig validation here.
Would a full default-state comparison be safer?

Fixed: isCustomGenesis now compares all 6 leaves of the default assertion state (added u64Vals[1], machineStatus, endHistoryRoot).


alec also noticed that setValidKeySet is doing a simulate transaction with the account.address, which fails nodes that dont support send_transaction (only send raw). So I've also updated those cases to use the full account, which contains the supplied priv key: https://github.com/OffchainLabs/arbitrum-chain-sdk/pull/705/changes/b96bae71732f2da9a78e20621585490b01f95f9a

douglance (approved):

PR #710, opened 2026-05-05
fix: aep example has broken L1 RPC and missing deps

godzillaba:
i had to make these changes to get the aep example working

### DIFF
--- a/examples/setup-aep-fee-router/index.ts
+++ b/examples/setup-aep-fee-router/index.ts
@@ -52,7 +52,7 @@ const chainOwner = privateKeyToAccount(sanitizePrivateKey(process.env.CHAIN_OWNE
 const parentChain = getParentChainFromId(Number(process.env.PARENT_CHAIN_ID));
 const parentChainPublicClient = createPublicClient({
   chain: parentChain,
-  transport: http(),
+  transport: http(process.env.PARENT_CHAIN_RPC),
 });
 
 // define chain config for the orbit chain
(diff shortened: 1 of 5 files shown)


examples/setup-aep-fee-router/index.ts
```diff
@@ -52,7 +52,7 @@ const chainOwner = privateKeyToAccount(sanitizePrivateKey(process.env.CHAIN_OWNE
 const parentChain = getParentChainFromId(Number(process.env.PARENT_CHAIN_ID));
 const parentChainPublicClient = createPublicClient({
   chain: parentChain,
-  transport: http(),
+  transport: http(process.env.PARENT_CHAIN_RPC),
```
alxdca, line 61: Are we fine with the fallback to the default rpc if the env var is not set ?

godzillaba: the default one is a dead cloudflare rpc.

personally i think it's fine, but defer to you guys

godzillaba: at least for L1 that's the default

alxdca: Let's add a warning like other examples do:
```
if (typeof process.env.PARENT_CHAIN_RPC === 'undefined' || process.env.PARENT_CHAIN_RPC === '') {
  console.warn(
    `Warning: you may encounter timeout errors while running the script with the default rpc endpoint. Please provide the "PARENT_CHAIN_RPC" environment variable instead.`,
  );
}
```

godzillaba: sounds good! c59399c149682cb577217b441b5f2242f240f061

alxdca (approved):

PR #713, opened 2026-05-19
feat: abi schema generation

yahgwai:
Generate Zod schemas for all contract ABIs and add per-contract CLI subcommands driven by those schemas.

  - Adds abi-to-zod-based schema generation: new pnpm generate:schemas script wired as postgenerate, so pnpm generate now emits a
  *.schemas.ts next to every contract under src/contracts/
  - Adds per-contract CLI subcommands (e.g. Rollup@v3.2, Rollup, ArbOwner, SequencerInbox). Each command is a discriminated union over
   function names, built generically from the ABI + the generated schemas
  - One command per contract; versioned contracts also get an unversioned entry pointing at the latest (e.g. Rollup duplicates
  Rollup@v3.2)
  - Read vs write is inferred from stateMutability. Reads run through readContract; writes through prepareTransactionRequest
  (unsigned, no broadcast) with optional upgradeExecutor wrapping via executeCall. value is only accepted on payable functions


douglance:
@yahgwai do we need to commit the schemas or could we just gen at build time? 

we have `scripts` and `scripting` folders. should they be combined? 

seems like we should discuss more about including the abi-to-zod lib.

yahgwai:
> do we need to commit the schemas or could we just gen at build time?

Could do yeah, but I figured it would be nice to inspect them easily, especially given abi-to-zod is quite new. We also have this convention already with the abis right. Can defo change this if you feel strongly about it though.
 
> we have `scripts` and `scripting` folders. should they be combined?

I think instead of merge we should probably rename `scripting` to `cli`? If so, I'll do that on a new branch

douglance (approved):

PR #718, opened 2026-06-09
feat: add ExpressLaneAuction (Timeboost) deployment

yahgwai:
- Add `deployProxyAdmin`: deploys an OpenZeppelin ProxyAdmin on the orbit chain (owned by the deployer).
- Add `deployExpressLaneAuction`: deploys the Timeboost ExpressLaneAuction implementation and an initialized TransparentUpgradeableProxy in front of it, given a ProxyAdmin.
- Expose both as CLI commands (`deployProxyAdmin`, `deployExpressLaneAuction`).
- Add `@arbitrum/nitro-contracts@^3.2.0` as the source of the contract ABIs/bytecode.
- CLI: resolve unrecognised orbit chain ids via a `defineChain` fallback (`withChildChainSign` + contract commands), so child-chain commands work on chains not in the registry.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -194,6 +194,9 @@ importers:
 
   src:
     dependencies:
+      '@arbitrum/nitro-contracts':
+        specifier: ^3.2.0
+        version: 3.2.0
       '@arbitrum/sdk':
         specifier: ^4.0.4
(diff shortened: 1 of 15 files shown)


spsjvc (approved):
LGTM

PR #728, opened 2026-07-13
fix(cli): resolve custom orbit chains for all cli commands

yahgwai:
- `findChain` now synthesizes a minimal chain for ids not in the registry instead of throwing, so CLI
  commands work against freshly deployed orbit chains.
- Fixes all call sites at once (ArbOS upgrade, allow-list, upgrade-executor, safe deploy, etc.) —
  previously only the deploy commands and generic contract calls were covered.
- Removes the redundant `findOrDefineChain`, consolidating to one resolver.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
       - dependency-name: '*'
         update-types:
           - 'version-update:semver-major'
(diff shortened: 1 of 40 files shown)


yahgwai:
Superseded by #731.

PR #731, opened 2026-07-15
feat: Support custom chains in the CLI

yahgwai:
- Add `deployWeth`, `deployRollupCreator` (v3.2), and `deployTokenBridgeCreator` SDK functions (viem ports of the upstream deploy scripts); retire the Docker-based token-bridge-creator helper
- Make `registerCustomParentChain` factory addresses optional and resolve registered custom chains first (they win over a same-id built-in)
- CLI: expose the three deploy commands, and register a custom parent chain from JSON in `createRollup`/`createTokenBridge` and the deploy workflows
- Add an end-to-end integration test: deploy the factories on the testnode L2, register it as a custom parent, and deploy a full rollup + token bridge through them
- Support custom child chains in the CLI: `findChain` synthesizes a minimal chain for an id not in the built-in list (e.g. a freshly deployed child arbitrum chain) instead of throwing, so commands and workflows can target it


src/scripting/viemTransforms.ts
```diff
@@ -1,28 +1,20 @@
-import { Chain, createPublicClient, createWalletClient, defineChain, http } from 'viem';
+import { Address, Chain, createPublicClient, createWalletClient, defineChain, http } from 'viem';
 import { privateKeyToAccount } from 'viem/accounts';
 import { sanitizePrivateKey } from '../utils/sanitizePrivateKey';
-import { chains, getCustomParentChains } from '../chains';
+import { chains, getCustomParentChains, registerCustomParentChain } from '../chains';
 
 export function findChain(chainId: number): Chain {
```
alxdca: What about naming this `findOrCreateChain` ?

alxdca (approved):

src/scripting/viemTransforms.ts
yahgwai: Went with findOrDefineChain — it's the existing name for the permissive version.

PR #740, opened 2026-08-04
test: remove legacy nitro-testnode integration tests

alxdca:
## Summary

- Remove the legacy `nitro-testnode` integration-test CI job while retaining the four maintained Arbitrum Testnode variants.
- Rename the local integration command from `test:integration:arbitrum-testnode` to `test:integration:local` and update the README instructions.
- Remove legacy Nitro container names and `/config` path fallbacks from the test helpers.

## Why / Context

The supported integration workflow now uses `litro` (arbitrum-testnode) images to run integration tests.

Removing the legacy Nitro testnode path avoids maintaining duplicate CI coverage and keeps the local documentation and test helpers aligned with the supported workflow.


yahgwai:
Why are these tests legacy now? Can you add something to the PR description

alxdca:
> Why are these tests legacy now? Can you add something to the PR description

Update the PR description but we currently have integrations running against both the legacy nitro-testnode and the new, much faster, litro (arbitrum-testnode). We've kept both in parallel to ensure that the new test image doesn't result in false positives/negatives but it's looking good so far. So we can remove the legacy tests for v1 but keep them on main until then.

yahgwai (approved):

PR #742, opened 2026-08-04
chore: update examples to use workspace chain-sdk

alxdca:
## Summary

- Add `@arbitrum/chain-sdk` as a `workspace:*` dependency to every example that did not already declare it.
- Update the pnpm lockfile so all 12 examples resolve the SDK through `link:../../src`.

## Why / Context

Examples import `@arbitrum/chain-sdk`, but most did not declare it directly. This could make them depend on incidental workspace hoisting or a published package instead of the SDK being developed locally.

Declaring the workspace dependency consistently ensures every example uses the current local SDK implementation.


yahgwai (approved):

yahgwai:
Out of interest, how did you find this issue?

alxdca:
> Out of interest, how did you find this issue?

By trying to build and run one of the examples.

PR #746, opened 2026-08-25
feat: support 3.2.0 upgrade action

alxdca:

* **#750**
* **#748**
* **#711**
* ➡️ **#746**
* `main`
</details></td></tr></table>

## Summary
- add APIs to prepare, deploy, execute, and verify Nitro contracts upgrade actions
- support the Nitro contracts 3.2.0 upgrade configuration
- add a reusable Forge script runner with Foundry validation and structured error details

## Why / Context
- enable SDK consumers to perform Nitro contract upgrades through prepared transaction requests or viem private-key accounts
- support offline or externally managed signing through the prepare/execute pattern


src/nitroContractsUpgrade.ts
```diff
@@ -0,0 +1,97 @@
...
+    Execute: require.resolve(
+      '@arbitrum/chain-actions/scripts/foundry/contract-upgrades/3.2.0/ExecuteNitroContracts3Point2Point0Upgrade.s.sol',
+    ),
+    Verify: require.resolve(
+      '@arbitrum/chain-actions/scripts/foundry/contract-upgrades/3.2.0/VerifyNitroContracts3Point2Point0Upgrade.s.sol',
+    ),
+  },
+} satisfies Record<NitroContractsUpgradeVersion, Record<UpgradeOperation, string>>;
+
+function runUpgradeScript(
```
yahgwai: Can we add this as to the cli, or should that be a separate PR?

src/nitroContractsUpgrade.ts
```diff
@@ -0,0 +1,97 @@
...
+  V3_2_0 = '3.2.0',
+}
+
+export type DeployNitroContractsUpgradeActionParameters = {
+  version: NitroContractsUpgradeVersion;
+  parentChainRpcUrl: string;
+  forgeArgs?: string[];
+};
+
+export type ExecuteNitroContractsUpgradeParameters = DeployNitroContractsUpgradeActionParameters & {
```
yahgwai: This will differ based on the specific upgrade right? so this would be `ExecuteNitroContracts3Point2UpgradeParameters` right? Same with deploy and verify i think

src/nitroContractsUpgrade.ts
alxdca: Yes, this is added in this PR: https://github.com/OffchainLabs/arbitrum-chain-sdk/pull/750

src/nitroContractsUpgrade.ts
alxdca: Why ? Do you expect future upgrade to require version specific parameters when sending the execute upgrade transaction ?

yahgwai: Yes, it's differed in the past, check out the args for the constructor and perform function of these contracts:
* https://github.com/OffchainLabs/arbitrum-chain-actions/blob/main/contracts/parent-chain/contract-upgrades/NitroContracts2Point1Point3UpgradeAction.sol
* https://github.com/OffchainLabs/arbitrum-chain-actions/blob/main/contracts/parent-chain/contract-upgrades/NitroContracts2Point1Point2UpgradeAction.sol

alxdca: Makes sense. I changed the execute type to be version specific.

src/nitroContractsUpgrade.ts
```diff
@@ -0,0 +1,353 @@
...
+
+const forgeDeploymentSchema = z.object({
+  transactions: z.tuple([
+    z.object({
+      transactionType: z.literal('CREATE2'),
+      contractAddress: addressSchema,
+      transaction: z.object({
+        from: addressSchema,
+        to: addressSchema,
+        value: hexSchema,
```
yahgwai: We have a big num schema elsewhere we could? There might also be hex and address schemas in the common

yahgwai (approved):

PR #748, opened 2026-09-01
feat: generate genesis

alxdca:

* **#750**
* ➡️ **#748**
* `main`
</details></td></tr></table>

## Summary

- Add a `generateGenesis` CLI command that creates an Orbit chain genesis configuration and derives its block hash and send root with Nitro's `genesis-generator`.
- Add a strict CLI schema for genesis parameters, including defaults for maximum code and init-code sizes.
- Install `@arbitrum/genesis-file-generator` in the chain SDK Docker image.
- Update the image to Nitro `v3.11.3-beb2108` and a Node 24 Bookworm runtime.

## Why / Context

This extends the Docker image introduced by the parent branch with the tooling required to generate chain genesis artifacts through the chain SDK CLI.

The implementation uses `@arbitrum/genesis-file-generator` to create the genesis configuration, then invokes Nitro's `genesis-generator` to calculate and validate the corresponding block hash and send root. 

Genesis generation is intentionally limited to the Docker image because the required package and Nitro binary are bundled there rather than added to the published SDK dependencies.

### DIFF
--- a/src/generateGenesis.ts
+++ b/src/generateGenesis.ts
@@ -1,0 +1,74 @@
+import { execFileSync } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { isHash } from 'viem';
+
+export type GenerateGenesisParameters = {
+  chainId: string;
+  isAnyTrust?: string;
+  arbosVersion: string;
+  chainOwner: string;
+  l1BaseFee: string;
+  loadDefaultPredeploys?: string;
+  enableNativeTokenSupply?: string;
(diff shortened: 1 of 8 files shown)


src/generateGenesis.ts
```diff
@@ -0,0 +1,74 @@
+import { execFileSync } from 'node:child_process';
```
yahgwai, line 1: Should we have an sdk function for generating genesis, and then just call the sdk function from in here? That would require someone to have installed the genesis generator on their machine I guess?

alxdca: We don't it exposed as an SDK function as it indeed requires the genesis-generator binary which we don't provide. It will be be callable using the CLI in docker image only.

yahgwai (approved):

PR #753, opened 2026-09-02
feat: support Robinhood Chain

godzillaba:
includes a bump of token-bridge-contracts to v1.2.5 in commit 9917641e094e11698e6255fd8aa414e5cd01958d, since that's the version deployed to RH chain.

i had to edit the `ContractConfig` type since I only deployed the latest version of the rollup creator, i'm not sure if that was the right thing to do though.


src/utils/getRollupCreatorAddress.ts
```diff
@@ -25,12 +25,22 @@ export function getRollupCreatorAddress<TChain extends Chain | undefined>(
...
-  return rollupCreatorAddress[parentChainId];
+  const address = rollupCreatorAddress[parentChainId];
+
+  if (typeof address === 'undefined') {
+    throw new Error(
+      `RollupCreator ${rollupCreatorVersion} is not deployed on parent chain ${parentChainId}`,
+    );
+  }
+
+  return address;
```
godzillaba: if we deprecate v2.1 then we can clean this up and remove the undefined check

wagmi.generate.ts
```diff
@@ -27,17 +29,28 @@ if (!apiKey) {
   throw new Error('Missing the ETHERSCAN_API_KEY environment variable!');
 }
 
+// address maps may omit parent chains where older contract versions are not deployed
 export type ContractConfig = {
   name: string;
   version?: string;
-  address: Record<ParentChainId, `0x${string}`> | `0x${string}`;
+  address: Partial<Record<ParentChainId, `0x${string}`>> | `0x${string}`;
+};
```
godzillaba, line 37: i had to change this type, not sure if that the right move

wagmi.generate.ts
```diff
@@ -27,17 +29,28 @@ if (!apiKey) {
...
-  if (chainId === arbitrumNova.id) {
-    const response = await fetch(
-      `https://arbitrum-nova.blockscout.com/api/v2/smart-contracts/${address}`,
-    );
+  const blockscoutUrl = blockscoutUrls[chainId];
+
+  if (blockscoutUrl) {
+    const response = await fetch(`${blockscoutUrl}/api/v2/smart-contracts/${address}`, {
+      // Some blockscout instances reject non-browser user agents
+      headers: { 'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36' },
```
godzillaba, line 52: specifically calling out this hack because it's kinda ugly

src/utils/getRollupCreatorAddress.ts
godzillaba: leaving deprecation for another time, removed comment in 893da4cb8159b72e8eddbcfbb6ec12e9aefd8b0d

spsjvc (approved):
LGTM

